PR #64, opened 2021-06-01
Feature ABC-202 add layout mobile to profile card

renaudaf:
- Added a new render() to profile card for mobile
- Added avatar-mobile-position


renaudaf:
The only thing missing is the font size for the different sizes.

jbverger (approved):

PR #68, opened 2021-06-03
ABC-213 :Create "Input Choice Set" component

Cedchillis:


src/modules/base/inputChoiceSet/__tests__/inputChoiceSet.test.js
```diff
@@ -174,10 +184,31 @@ describe('Checkbox Group', () => {
...
+    // options with icons
+    it('Input choice set options with icons', () => {
+        const element = createElement('base-input-choice-set', {
+            is: InputChoiceSet
+        });
+        document.body.appendChild(element);
+
+        element.options = optionsWithIcon;
+
+        return Promise.resolve().then(() => {
```
nawel-aa, line 196: I know these tests already existed, but they will always pass, so they need to be changed.
`options[index]` and `option` are exactly the same object. Instead we should test if DOM elements receive the label/icon of `option`. 🙂

src/modules/base/inputChoiceSet/inputChoiceSet.css
```diff
@@ -0,0 +1,46 @@
+:host {
```
nawel-aa, line 1: I don't think you need any of the `!important`, so it would be better to remove them.

src/modules/base/inputChoiceSet/inputChoiceSet.css
```diff
@@ -0,0 +1,46 @@
...
+    display: block;
+}
+
+.slds-checkbox.horizontal{
```
nawel-aa, line 5: You can probably apply `.slds-grid` on your `computedButtonClass` instead of having to use `display: inline-flex` on all options (I'll let you confirm, maybe I'm wrong 🙂 ).

src/modules/base/inputChoiceSet/inputChoiceSet.css
```diff
@@ -0,0 +1,46 @@
...
+.slds-checkbox_button__label.vertical{
+    width: 100%;
+    border-radius: 0 !important;
+}
+
+.slds-checkbox_button__label.horizontal{
+    height: 100%;
+}
+
+.block{
```
nawel-aa, line 26: I don't think the class `block` is used.

src/modules/base/inputChoiceSet/inputChoiceSet.js
```diff
@@ -13,28 +13,33 @@ import {
     VARIANT
 } from 'c/inputUtils';
 import { classSet } from 'c/utils';
+import InputChoiceOption from './inputChoiceOption'
+import { check } from 'prettier';
```
nawel-aa, line 17: This import is not used.

nawel-aa (changes requested):

PR #69, opened 2021-06-03
[feat](ABC-223): Added description, icons and infos

Abderraouf99:


src/modules/base/list/list.css
```diff
@@ -39,9 +35,8 @@
     background-color: rgba(255, 255, 255, 0.6);
 }
 
-.sortable-item_moved-up {
-    transform: translateY(calc(0px - var(--item-height)));
-}
-.sortable-item_moved-down {
-    transform: translateY(var(--item-height));
+.description-text {
```
nawel-aa, line 38: You can use the class `slds-truncate` instead of creating `description-text`

src/modules/base/list/list.js
```diff
@@ -43,6 +44,14 @@ export default class List extends LightningElement {
     set items(proxy) {
         this._items = normalizeArray(proxy);
         this.computedItems = JSON.parse(JSON.stringify(this._items));
+        this.computedItems.forEach((item) => {
+            if (item.infos === undefined) {
+                item.infos = [];
```
nawel-aa, line 49: You can use `normalizeArray()` instead. This way it will work whatever invalid value is passed (undefined, string, null, etc.).
```
item.infos = normalizeArray(item.infos);
item.icons = normalizeArray(item.icons);
```

nawel-aa (changes requested):
Hi Abderraouf!
I only noticed two small normalization issues where you could use the SLDS existing tools instead of creating new ones. :)

nawel-aa (approved):

PR #79, opened 2021-06-10
(feat) (abc 201) input counter type attribute

AlainRusnov:
Added ‘type’ property (string): 

- ‘number’, ‘currency’, ‘percent’ ( ‘number’ default and ‘percent’ not percent-fixed )

Added ‘input-step’ property (number):  

- Granularity of the value - precision of significant numbers ( specified as a positive floating point number. ex: 0.01 formats the value to 2 digits after the decimal  ). Default: 0.01 in storybook

Added handler for loss of precision when inc/dec value. Passes step validation.

When Max-Min are called on and there is a validation error, the value looses its formatting ( e.g. currency or % sign ), until it meets the criteria.

Tests : type, input-step, precision


src/modules/base/inputCounter/inputCounter.js
```diff
@@ -94,27 +94,25 @@ export default class InputCounter extends LightningElement {
...
-        this.checkValidity();
     }
 
     @api get min() {
         return this._min;
     }
 
     set min(min) {
-        this._min = typeof min === 'number' ? this.handlePrecision(min) : '';
+        this._min = typeof min === 'number' ? min : '';
```
jbverger, line 107: If _min is a number, default value should be null?

src/modules/base/inputCounter/inputCounter.html
```diff
@@ -27,33 +59,27 @@
...
                 required={required}
                 type="number"
-                placeholder="0"
+                formatter={type}
                 class={inputClass}
                 onchange={handlerChange}
                 onfocus={handlerFocus}
                 onblur={handlerBlur}
-            ></lightning-input>
+                autocomplete="off"
```
jbverger, line 81: autocomplete is not available for type==number.
_Use this attribute with email, search, tel, text, and url input types only_ https://developer.salesforce.com/docs/component-library/bundle/lightning-input/specification

src/modules/base/inputCounter/inputCounter.js
```diff
@@ -51,11 +94,39 @@ export default class InputCounter extends LightningElement {
...
+    }
+    
+    @api get fractionDigits() {
+        return this._fractionDigits;
+    }
+
+    set fractionDigits(digits) {
+        const baseFractionValue = 1;
+        this._fractionDigits = typeof digits === 'number' ? +('0.' + baseFractionValue.toString().padStart(digits, '0')) 
+        : this.step.toString().includes('.' || ',') ? 0.1
```
jbverger, line 125: No need to manage that case. If the step include decimal, the user need to specify a fractionDigits. Same as lightning-input, the user need to specify a step for decimal value.

jbverger (changes requested):

src/modules/base/inputCounter/inputCounter.css
```diff
@@ -43,6 +75,10 @@
     padding-top: 5px;
 }
 
+:host(.avonni-flex-container) > .slds-form-element__help {
+    padding-top: 5px;
```
nawel-aa, line 79: You can probably use `.slds-p-top_xx-small` on the element instead of a custom CSS rule.

src/modules/base/inputCounter/inputCounter.html
nawel-aa: Screen Shot 2021-06-17 at 11 15 51 AM

But when we remove it, the browser adds a form autocomplete. 😕

src/modules/base/inputCounter/inputCounter.js
```diff
@@ -162,51 +244,103 @@ export default class InputCounter extends LightningElement {
...
-        this.template.querySelector('lightning-input').focus();
-    }
-
-    @api
-    blur() {
-        this.template.querySelector('lightning-input').blur();
-    }
+            
+        decrementValue() {
+            this.max? this.handlePrecision(this.max) : null;
```
nawel-aa, line 249: We try to pass eslint validation as much as possible, even if it means writing less cool looking code. 😄 
So instead of ternary operators, you should use `if (this.max) this.handlePrecision(this.max);`

src/modules/base/inputCounter/inputCounter.js
```diff
@@ -162,51 +244,103 @@ export default class InputCounter extends LightningElement {
...
+            this.min? this.handlePrecision(this.min) : null;
+            this.step = this.handlePrecision(this.step);
+            
+            const previousValue = this.value? +(this.value) : null;
+            
+            if (this.value !== undefined && !isNaN(this.value)) {
```
nawel-aa, line 255: You can probably just check if `!isNaN(this.value)`, since `!isNaN(undefined)` will return false?

src/modules/base/inputCounter/inputCounter.js
```diff
@@ -162,51 +244,103 @@ export default class InputCounter extends LightningElement {
...
-
-    @api
-    blur() {
-        this.template.querySelector('lightning-input').blur();
-    }
+            
+        decrementValue() {
+            this.max? this.handlePrecision(this.max) : null;
+            this.min? this.handlePrecision(this.min) : null;
+            this.step = this.handlePrecision(this.step);
```
nawel-aa, line 251: Same as above, unless there's a specific reason for not doing it, you should assign new values to `this._step` instead of `this.step`.

src/modules/base/inputCounter/inputCounter.js
```diff
@@ -162,51 +244,103 @@ export default class InputCounter extends LightningElement {
...
+            
+            const previousValue = this.value? +(this.value) : null;
+            
+            if (this.value !== undefined && !isNaN(this.value)) {
+                this.value = Number(this.value) - Number(this.step);           
+                if (this.min || this.min === 0) {
```
nawel-aa, line 257: Is there a value in having `this.value = this.value` ? If not, the block could probably be changed into:
```
if ((this.min || this.min === 0) && (this.value - this.step < this.min)) {
  this._value = this.min;
}
```

src/modules/base/inputCounter/inputCounter.js
```diff
@@ -215,38 +349,109 @@ export default class InputCounter extends LightningElement {
...
+    @api
+    checkValidity() {
+        return this._constraint.checkValidity();
+    }
+
+    @api
+    reportValidity() {
+        let helpMsg = '';
+
+        let validationMsg = this._constraint.reportValidity((message) => {
```
nawel-aa, line 413: I am not sure why you use helpMsg. You can probably just return the result of `this._constraint.reportValidity(...)` and assign the message to `this.helpMessage` inside of it (same as what have been done in dualListbox).

nawel-aa (changes requested):

jbverger (approved):

PR #93, opened 2021-06-17
Feature ABC-261 implement getListHeight() in dual listbox

renaudaf:


src/modules/base/dualListbox/dualListbox.html
```diff
@@ -92,7 +92,7 @@
                             >
                                 <li
                                     key={option.value}
-                                    role="presentation"
```
nawel-aa, line 95: The role needs to be one of the accepted [ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) for accessibility. Since we are in a listbox, the idea of using `role="presentation"` is to indicate that even though it's an `<li>` tag, it's not the actual list item. The list item is the `<div>` below that has the `role="option"`.
You can add a class or a `data-role="source"` if you need to select the items in your Javascript. :)

src/modules/base/dualListbox/dualListbox.js
```diff
@@ -557,137 +546,23 @@ export default class DualListbox extends LightningElement {
...
-                        this._selectedHasDescription
-                    ) +
-                    42.3 *
-                        (this._maxVisibleOptions -
-                            this.computedSelectedList.length);
-            }
-        }
-        return this._selectedBoxHeight;
-    }
+        this._sourceBoxHeight = getListHeight(
```
nawel-aa, line 557: I may have misunderstood what the code is doing, but instead of repeating the searchEngine calculation in `get sourceHeight()` and `get selectedHeight()`, maybe you could just add the search engine element to the `sourceOptions` array if there is a search box, so that you're sure that it's always included in the source height? 
This way, instead of having `get sourceHeight()` and `get selectedHeight()`, you could have only one get `listHeight`, that would always return the biggest number of the two heights?

nawel-aa (changes requested):

nawel-aa (approved):

PR #94, opened 2021-06-18
(feat) (abc 243) list component actions display

AlainRusnov:
List actions : 
- 1 action no label : button-icon
- 1 action label : button
- multi actions : button menu

### DIFF
--- a/src/modules/base/list/list.html
+++ b/src/modules/base/list/list.html
@@ -125,7 +125,10 @@
                         icon-name={sortableIconName}
                         size="x-small"
                     ></lightning-icon>
-                    <div class="slds-media slds-media_small">
+                    <div
+                        if:true={actions.length}
+                        class="slds-media slds-media_small"
+                    >
                         <lightning-button-menu
                             if:true={multiActions}
                             class="slds-m-right_none"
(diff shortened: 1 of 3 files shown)


jbverger (changes requested):

src/modules/base/list/list.html
```diff
@@ -125,23 +125,52 @@ <h3>{item.label}</h3>
...
+                                <lightning-menu-item
+                                    key={action.name}
+                                    value={action.name}
+                                    label={action.label}
+                                    disabled={action.disabled}
+                                    icon-name={action.iconName}
+                                ></lightning-menu-item>
+                            </template>
+                        </lightning-button-menu>
+                        <template if:false={multiActions}>
```
nawel-aa, line 149: You should not have two `template if` one inside each other, it's better to keep your logic inside your Javascript file (see [guidelines](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019781/HTML+Templates)).
Maybe you could use only`if:false={multiActions}`, and add a `if:true={actions.length}` to your `<div>` line 128, to handle the case of empty actions?

nawel-aa (changes requested):
There's a small HTML syntax problem, but most importantly, there are no tests for the actions attribute.

src/modules/base/list/list.js
```diff
@@ -130,6 +131,14 @@ export default class List extends LightningElement {
         this._hasActions = true;
     }
 
+    get singleAction() {
+        return this.computedActions[0];
+    }
+
+    get multiActions() {
```
jbverger, line 138: Rename multiActions to hasMultipleActions

src/modules/base/list/list.js
```diff
@@ -130,6 +131,14 @@ export default class List extends LightningElement {
         this._hasActions = true;
     }
 
+    get singleAction() {
```
jbverger, line 134: Rename singleAction to firstAction.

jbverger (approved):

PR #96, opened 2021-06-22
(feat) (abc 241) seperator

AlainRusnov:
Seperator component : 
- Label ( string )
- Icon-name ( string )
- orientation ( vertical/horizontal - string) default horizontal
- align-content ( start, center, end - string) default center
- icon-size ( xx-small, x-small, small, medium, large - string ) default medium
- icon-position ( left, right - string) default left

Added storybookwrapper to set a surrounding container height of 400px.


src/modules/base/separator/separator.js
```diff
@@ -0,0 +1,162 @@
...
+        return classSet(
+            'avonni-separator_content slds-grid slds-vertical slds-grid_vertical-align-center slds-grid_align-center slds-m-around_small'
+        )
+            .add({
+                'slds-grid_reverse': this.iconPosition === 'right'
+            })
+            .toString();
+    }
+
+    get computedIconMargin() {
```
jbverger, line 152: Can you rename to computedIconClass? Thanks,

PR #221, opened 2021-11-25
Fix ABC-420 Fixed issue when no href in item attribute

renaudaf:


renaudaf:
- Creation of primitive carousel item
- Added tests for primitive carousel item

jbverger (approved):

PR #266, opened 2022-02-25
[feat] (ABC-489): Add and improve list styling hooks

lgilmet:
- --avonni-list-item-color-background-hover-sortable ✅
- --avonni-list-item-color-background-sortable ✅
- --avonni-list-item-spacing-block-between ✅
- --avonni-list-item-spacing-block-start ✅
- --avonni-list-item-spacing-block-end ✅
- --avonni-list-item-spacing-inline-start ✅
- --avonni-list-item-spacing-inline-end ✅
- --avonni-list-item-cursor ✅
- --avonni-list-item-cursor-sortable ✅

### DIFF
--- a/src/modules/base/list/__docs__/list.jsdoc.js
+++ b/src/modules/base/list/__docs__/list.jsdoc.js
@@ -153,7 +153,7 @@
  */
 /**
  * @memberof stylingHooks
- * @name --avonni-list-item-color-background-hover-sortable
+ * @name --avonni-list-item-color-background-sortable-hover
  * @default --avonni-list-item-color-background-hover, #f3f2f2
  * @type spacing
  */
@@ -164,32 +164,32 @@
  */
 /**
  * @memberof stylingHooks
- * @name --avonni-list-item-block-start
+ * @name --avonni-list-item-spacing-block-start
  * @type sizing
  */
 /**
  * @memberof stylingHooks
- * @name --avonni-list-item-block-end
+ * @name --avonni-list-item-spacing-block-end
  * @type sizing
  */
 /**
  * @memberof stylingHooks
- * @name --avonni-list-item-inline-start
+ * @name --avonni-list-item-spacing-inline-start
  * @type sizing
  */
 /**
  * @memberof stylingHooks
- * @name --avonni-list-item-inline-end
+ * @name --avonni-list-item-spacing-inline-end
  * @type sizing
  */
 /**
  * @memberof stylingHooks
  * @name --avonni-list-item-cursor-sortable
  * @default --avonni-list-item-cursor, grab
- * @type styling
+ * @type string
  */
 /**
  * @memberof stylingHooks
  * @name --avonni-list-item-cursor
- * @type styling
+ * @type string
  */
--- a/src/modules/base/list/__docs__/list.jsdoc.js
+++ b/src/modules/base/list/__docs__/list.jsdoc.js
@@ -189,6 +189,6 @@
  */
 /**
  * @memberof stylingHooks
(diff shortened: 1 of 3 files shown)


src/modules/base/list/list.css
```diff
@@ -33,6 +33,14 @@
...
     height: fit-content;
     background-color: var(--avonni-list-item-color-background, transparent);
+    padding-top: var(--avonni-list-item-block-start);
+    padding-bottom: var(--avonni-list-item-block-end);
+    padding-left: var(--avonni-list-item-inline-start);
+    padding-right: var(--avonni-list-item-inline-end);
+}
+
+.avonni-list__item:not(:last-child) {
+    margin-bottom: var(--avonni-list-item-spacing);
```
nawel-aa, line 43: Maybe `--avonni-list-item-spacing-block-between`, to follow the block/inline convention?

src/modules/base/list/list.js
```diff
@@ -353,7 +353,14 @@ export default class List extends LightningElement {
                 'avonni-list__item-sortable': this.sortable,
                 'avonni-list__item-expanded': this._hasActions,
                 'slds-p-vertical_x-small': !this._divider,
-                'slds-p-horizontal_none': this._divider === 'top' || 'bottom',
+                /**
```
nawel-aa, line 356: I agree that you should remove this class.

src/modules/base/list/__docs__/list.jsdoc.js
```diff
@@ -146,3 +157,38 @@
  * @default 0.25rem
  * @type radius
  */
+/**
+ * @memberof stylingHooks
+ * @name --avonni-list-item-spacing
+ * @type sizing
```
nawel-aa, line 163: The type would be spacing when it's margin/padding. Sizing is used for widths and heights.

nawel-aa (changes requested):
See also my comments on Jira's ABC-490 and ABC-489.

src/modules/base/list/__docs__/list.jsdoc.js
```diff
@@ -146,3 +151,45 @@
...
+/**
+ * @memberof stylingHooks
+ * @name --avonni-list-item-inline-end
+ * @type sizing
+ */
+/**
+ * @memberof stylingHooks
+ * @name --avonni-list-item-cursor-sortable
+ * @default --avonni-list-item-cursor, grab
+ * @type styling
```
nawel-aa, line 189: I think the type is supposed to be a [CSS data type](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types). So I would go with "string" for the cursors, since nothing else seems to fit.

src/modules/base/list/__docs__/list.jsdoc.js
```diff
@@ -146,3 +151,45 @@
  * @default 0.25rem
  * @type radius
  */
+/**
+ * @memberof stylingHooks
+ * @name --avonni-list-item-color-background-hover-sortable
```
nawel-aa, line 156: After talking with Renaud, I think that since "sortable" can be used with "hover", it should be considered an attribute and not a state. So according to the naming convention, it would be `--avonni-list-item-color-background-sortable-hover`.

nawel-aa (changes requested):
Works like a charm, but after a second careful review, there are some naming and doc changes to make. 🙂

lgilmet:
Hooks renamed and tested.

lgilmet:
Hook type spacing changed to dimension

nawel-aa (approved):

PR #267, opened 2022-02-25
Feature abc 471 dual list box searching case

lgilmet:
Fix search in dual listbox so that you can lowercase and uppercase both return results. 
Fix rangeOverflow error message when max is not set. 
Default max now set to undefined instead of empty string.

### DIFF
--- a/src/modules/base/dualListbox/dualListbox.js
+++ b/src/modules/base/dualListbox/dualListbox.js
@@ -43,7 +43,6 @@ import { FieldConstraintApi, InteractingState } from 'c/inputUtils';
 import { handleKeyDownOnOption } from './keyboard';
 
 const DEFAULT_MIN = 0;
-const DEFAULT_MAX = undefined;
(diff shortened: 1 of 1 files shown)


src/modules/base/dualListbox/dualListbox.js
```diff
@@ -476,7 +477,7 @@ export default class DualListbox extends LightningElement {
     }
 
     set max(value) {
-        const number = isNaN(parseInt(value, 10)) ? '' : value;
+        const number = isNaN(parseInt(value, 10)) ? DEFAULT_MAX : value;
```
nawel-aa, line 480: Since the default is `undefined`, we don't have to use a `DEFAULT_MAX` variable.
I would also suggest that we always `parseInt()` the value, so we are sure the value of `max` is always a number and is safe to add, compare, etc. Same with `min` and `maxVisibleOptions`.

nawel-aa (commented):
Works well, just a comment on convention.

PR #273, opened 2022-03-09
[fix][docs] Correct hook types, no ',' in @type

lgilmet:
Reverted all to previous versions. Corrected on release.


src/modules/base/list/__docs__/list.jsdoc.js
```diff
@@ -42,25 +42,25 @@
...
  * @name --avonni-list-header-font-style
  * @default normal
- * @type string
+ * @type font
  */
 /**
  * @memberof stylingHooks
  * @name --avonni-list-header-font-weight
  * @default 400
- * @type string, number
```
lgilmet, line 57: Didn't know I did that.

nawel-aa (approved):

PR #276, opened 2022-03-10
Feature abc 479 add is joined to page header

lgilmet:
Added storybook controls and implemented isJoined to add slds-page-header_joined. Works on vertical too.

Removed slds classes from title slot.

### DIFF
--- a/src/modules/base/pageHeader/__examples__/base/base.html
+++ b/src/modules/base/pageHeader/__examples__/base/base.html
@@ -4,7 +4,6 @@
         icon-name="standard:opportunity"
         info="Info"
         variant="base"
-        is-joined={isJoined}
     >
         <div slot="title">Slot Title</div>
         <div slot="label">Slot Label</div>
--- a/src/modules/base/pageHeader/__examples__/base/base.html
+++ b/src/modules/base/pageHeader/__examples__/base/base.html
@@ -4,6 +4,7 @@
         icon-name="standard:opportunity"
         info="Info"
         variant="base"
+        is-joined={isJoined}
     >
         <div slot="title">Slot Title</div>
         <div slot="label">Slot Label</div>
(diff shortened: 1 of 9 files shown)


src/modules/base/pageHeader/__examples__/base/base.html
```diff
@@ -4,6 +4,7 @@
         icon-name="standard:opportunity"
         info="Info"
         variant="base"
+        is-joined={isJoined}
```
nawel-aa, line 7: The base example is normally as uncustomized as possible, but you can add `is-joined` to one of the other examples (`objectHome`, for example), and change it in the stories too.

src/modules/base/pageHeader/pageHeader.html
```diff
@@ -65,11 +65,16 @@
                             <div class="slds-page-header__name-title">
                                 <template if:true={isBaseVariant}>
                                     <span
-                                        class="slds-page-header__title slds-truncate"
+                                        class="
+                                            slds-page-header__title
+                                            slds-truncate
+                                        "
                                     >
                                         <template if:true={hasStringTitle}>
```
nawel-aa, line 73: The `if:true` can be placed directly on the new spans, we don't need the `<template>` tags anymore.

nawel-aa (changes requested):
The broken tests should be fixed, and new ones should be added for the new attribute.
And don't forget to keep Confluence up to date! 🙂

src/modules/base/pageHeader/__examples__/base/base.html
lgilmet: I just forgot it there after trying to make storybook work.

nawel-aa (approved):

lgilmet:
Doc
- [ ] https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/1766457432/Page+Header

PR #286, opened 2022-03-25
[feat] (ABC-513): Add fallback bullet for omitted icons

lgilmet:
Add isActive to item. If item has no icon it has a gray bullet. If also isActive, item has a blue bullet.

Screen Shot 2022-03-31 at 11 19 25 AM

https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/1620082724/Activity+Timeline
https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/1621819399/Activity+Timeline+Item

### DIFF
--- a/src/modules/base/activityTimeline/__examples__/progressIndicator/progressIndicator.html
+++ b/src/modules/base/activityTimeline/__examples__/progressIndicator/progressIndicator.html
@@ -1,0 +1,10 @@
+<template>
+    <avonni-activity-timeline
+        items={items}
+        actions={actions}
+        collapsible="true"
+        title="Activity Timeline"
+        icon-name="standard:timesheet_entry"
+        variant="progress-indicator"
+    ></avonni-activity-timeline>
+</template>
--- a/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
+++ b/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
@@ -58,7 +58,6 @@ describe('Activity Timeline', () => {
         expect(element.items).toMatchObject([]);
         expect(element.sortedDirection).toBe('desc');
         expect(element.title).toBeUndefined();
-        expect(element.variant).toBe('base');
     });
 
     /* ----- ATTRIBUTES ----- */
@@ -157,7 +156,7 @@ describe('Activity Timeline', () => {
     });
 
     // variant
-    it('Activity timeline: variant progress-indicator', () => {
+    it('Activity timeline: item iterator', () => {
         const ITEM = [
             {
                 name: 'item1',
@@ -165,7 +164,6 @@ describe('Activity Timeline', () => {
                 description: 'You logged a call with Adam Chan',
                 href: '#',
                 datetimeValue: 1653141600000,
-                iconName: 'standard:log_a_call',
                 fields: [
                     {
                         label: 'Name',
@@ -227,7 +225,6 @@ describe('Activity Timeline', () => {
         ];
 
         element.items = ITEM;
-        element.variant = 'progress-indicator';
 
         return Promise.resolve().then(() => {
             const items = element.shadowRoot.querySelectorAll(
@@ -235,9 +232,7 @@ describe('Activity Timeline', () => {
             );
 
             expect(items).toHaveLength(2)
-            expect(items[0].variant).toEqual('progress-indicator');
             expect(items[0].index).toBe(0);
-            expect(items[1].variant).toEqual('progress-indicator');
             expect(items[1].index).toBe(1);
         });
     });
@@ -273,7 +268,7 @@ describe('Activity Timeline', () => {
     });
 
     // items
-    it('Activity Timeline: items, base variant', () => {
+    it('Activity Timeline: items', () => {
         const ITEM = [
             {
                 name: 'item1',
@@ -363,7 +358,6 @@ describe('Activity Timeline', () => {
                 expect(item.loadingStateAlternativeText).toBe(
                     ITEM[index].loadingStateAlternativeText
                 );
-                expect(item.variant).toEqual('base');
                 expect(item.closed).toBe(ITEM[index].closed || false);
                 expect(item.buttonLabel).toBe(ITEM[index].buttonLabel);
                 expect(item.buttonIconName).toBe(ITEM[index].buttonIconName);
--- a/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
+++ b/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
@@ -232,8 +232,6 @@ describe('Activity Timeline', () => {
             );
 
             expect(items).toHaveLength(2)
-            expect(items[0].index).toBe(0);
-            expect(items[1].index).toBe(1);
         });
     });
 
--- a/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
+++ b/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
@@ -58,6 +58,7 @@ describe('Activity Timeline', () => {
         expect(element.items).toMatchObject([]);
         expect(element.sortedDirection).toBe('desc');
         expect(element.title).toBeUndefined();
+        expect(element.variant).toBe('base');
     });
 
     /* ----- ATTRIBUTES ----- */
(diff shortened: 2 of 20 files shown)


src/modules/base/primitiveActivityTimelineItem/__tests__/primitiveActivityTimelineItem.test.js
```diff
@@ -93,12 +93,13 @@ describe('Primitive Activity Timeline Item', () => {
         expect(element.buttonIconPosition).toBe('left');
         expect(element.buttonVariant).toBe('neutral');
         expect(element.buttonDisabled).toBeFalsy();
+        expect(element.index).toBeFalsy();
```
nawel-aa, line 96: We usually use `toBeFalsy()` for booleans. Here it could be `toBeUndefined()`, or maybe even `toBe(0)`, since it's supposed to be a number?

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.css
```diff
@@ -30,18 +30,51 @@
...
 .avonni-activity-timeline-item-no-fields_margin {
     margin-left: 24px;
 }
+
+.avonni-activity-timeline-item-checkbox {
+    /* Stop label retracting under checkbox and push checkbox up */
+    min-width: 22px;
+    max-width: 22px;
+    padding-bottom: 3px;
+}
```
nawel-aa, line 80: This is a complicated solution for a simpler problem, you can fix it by fixing your truncation problem at the same time. 🙂

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.html
```diff
@@ -80,15 +82,18 @@
...
                                 data-element-id="lightning-input-checkbox"
                                 onchange={handleCheck}
                             >
                             </lightning-input>
                         </template>
                         <template if:true={href}>
-                            <a href={href}>
-                                <h3>{title}</h3>
-                            </a>
+                            <span class="slds-trucate">
```
nawel-aa, line 92: Why adding a `<span>` around the `<a>`? If anything, you could make it even simpler by adding the `if:true` directly on the <a>.
Also, I trust that you can figure out how to make `slds-truncate` work. 😊 A hint is that one of the main things it does, is adding a `max-width` of 100%. But a max-width depends on the parent also having some sort of width...

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.js
```diff
@@ -290,6 +291,30 @@ export default class PrimitiveActivityTimelineItem extends LightningElement {
         this._isLoading = normalizeBoolean(value);
     }
 
+    /**
+     * Check if variant is progress-indicator.
+     *
+     * @type {boolean}
+     */
+    get hasNoIcon() {
```
nawel-aa, line 299: You don't need this getter, you can do it directly in your HTML.

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.js
```diff
@@ -290,6 +291,30 @@ export default class PrimitiveActivityTimelineItem extends LightningElement {
...
+     * @type {number}
+     * @public
+     */
+    @api
+    get index() {
+        return this._index;
+    }
+
+    set index(value) {
+        this._index = value;
```
nawel-aa, line 315: There are four reasons that come to my mind, for which we'd want to use a get/set instead of just having a `@api nameOfProperty`:
- There's a default value that you want to use if the given one is empty or invalid.
- You need to normalize the value.
- You need to do something else every time a new value is passed (for example, call a method).
- You will need to update the property yourself, so you need a private property to follow the [no api reassignment](https://github.com/salesforce/eslint-plugin-lwc/blob/master/docs/rules/no-api-reassignments.md) rule.

You can decide with this info if your index should be adapted to fit in one of these categories, or if you should just have an `@api index` at the top of your file. 😊

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.js
```diff
@@ -431,7 +470,7 @@ export default class PrimitiveActivityTimelineItem extends LightningElement {
      * @returns {string} line background color
      */
     setLineColor() {
-        const icon = this.template.querySelector('lightning-icon');
+        const icon = this.template.querySelector('.slds-timeline__icon');
```
nawel-aa, line 473: I know the old code was wrong too, but this query selector should be changed to use a `data-element-id` attribute.

nawel-aa (changes requested):
Works well. 👍 My main request is about the truncation that I know can be fixed. 😊

src/modules/base/activityTimeline/activityTimeline.html
```diff
@@ -53,6 +53,7 @@ <h2 class="slds-section__title">{title}</h2>
             <template for:each={sortedItems} for:item="item">
```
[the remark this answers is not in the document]
lgilmet: For those with a section, the first of every section becomes first, but have to mark only the very first item :(
Screen Shot 2022-03-30 at 10 51 43 AM

src/modules/base/activityTimeline/__examples__/progressIndicator/progressIndicator.html
```diff
@@ -0,0 +1,10 @@
+<template>
+    <avonni-activity-timeline
+        items={items}
+        actions={actions}
+        collapsible="true"
```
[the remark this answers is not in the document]
lgilmet: true true true

src/modules/base/primitiveActivityTimelineItem/__tests__/primitiveActivityTimelineItem.test.js
lgilmet: toBeUndefined, there is no default.

src/modules/base/activityTimeline/activityTimeline.html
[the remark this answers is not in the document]
nawel-aa: Silly me! 😃
You could still have a `groupIndex` attribute and an `index` attributes that should both be equal to 0, but your way is cleaner! 👌 You do have to reset your `_iterator` property every time the items are updated, though.

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.html
lgilmet: slds-truncate doesnt wanna work in the item header. it's all flex 😢.

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.js
lgilmet: touché

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.html
lgilmet: made it work 😌

lgilmet: This
```
<!-- this div.slds-has-flexi-truncate does the magic -->
<div class="slds-has-flexi-truncate">
    <h3 class="slds-truncate">
        {title}
    </h3>
</div>
```

src/modules/base/activityTimeline/activityTimeline.html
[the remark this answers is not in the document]
lgilmet: 😤

nawel-aa (approved):
🎉

src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
```diff
@@ -232,8 +232,6 @@ describe('Activity Timeline', () => {
             );
 
             expect(items).toHaveLength(2)
-            expect(items[0].index).toBe(0);
```
nawel-aa: I think this test can be removed. But you should add a line for `isActive`, in the test named "Activity Timeline: items".

nawel-aa (commented):
Works well! 👍 I only have a small suggestion about the tests. :)

src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
lgilmet: great!

nawel-aa (approved):

PR #301, opened 2022-04-20
Feature abc 522 add date range accessibility

lgilmet:
Add tabIndex='0' on the first of each month, otherwise focus can't enter calendar.

Input date range:
Return focus to input icon if 'escape', or on input field if 'selection', do nothing if click outside
If click input icon, focus active date or today.


src/modules/base/calendar/calendar.js
```diff
@@ -544,12 +580,16 @@ export default class Calendar extends LightningElement {
...
+                    // why does a disabled day have an adjacent month class?
+                    // dateClass = 'slds-day_adjacent-month';
+
+                    dateClass = '';
                     dayClass = 'avonni-calendar__disabled-cell slds-day';
                 } else {
                     fullDate = time;
                 }
 
+                fullDate = time;
```
nawel-aa, line 592: If you're going to set `fullDate` to `time` anyway, you don't need to have a variable. You can just set it in the object you're pushing into `weekData`.

src/modules/base/calendar/calendar.js
```diff
@@ -889,7 +933,8 @@ export default class Calendar extends LightningElement {
         this.dispatchEvent(
             new CustomEvent('change', {
                 detail: {
-                    value: this.normalizedValue
+                    value: this.normalizedValue,
+                    clickedDate: this._clickedDate
```
nawel-aa, line 937: For more consistency, it would be better if the `clickedDate` was an ISO string, instead of a date object.
Also, don't forget to update the JSDoc comment just above, and the Confluence documentation.

src/modules/base/calendar/calendar.js
```diff
@@ -914,6 +963,36 @@ export default class Calendar extends LightningElement {
         );
     }
 
+    /**
+     * Private focus out handler.
+     */
+    handleFocusOut() {
```
nawel-aa, line 969: I'm not sure why you need to implement this in the calendar. It seems like if the parent component needs it, it could do it itself around the calendar, since the advantage of `focusin` and `focusout` is that they bubble up. Also, `closingMethod` really only applies to the case where the calendar is inside a popover, but it's a component in itself that doesn't "close".

src/modules/base/calendar/calendar.js
```diff
@@ -1021,6 +1100,171 @@ export default class Calendar extends LightningElement {
...
+                this.handleFocusOut();
+                break;
+            default:
+                break;
+        }
+    }
+
+    /**
+     * Focus roving
+     * direction is number of days (+ : after, - : before)
```
nawel-aa, line 1160: The method's arguments should be documented through the [correct syntax](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#Allowed-block-tags-for-methods) (`@param {number} Description of the argument`). Don't hesitate to add the [JSDoc snippets](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#VSCode-snippet-configuration) to your VSCode, or make your own.

src/modules/base/calendar/calendar.js
```diff
@@ -1021,6 +1100,171 @@ export default class Calendar extends LightningElement {
...
+                                initialFocusDate.getDay()
+                        );
+                    }
+                    break;
+                default:
+                    console.log('Accessibility issue changing date');
+            }
+
+            if (nextDate) {
+                let currentDate = this.date.getTime();
```
nawel-aa, line 1206: All of the next variables could be constants. Always set your variables to a `const` first, and **then** update them to a `let` if you realize you need to edit them.

nawel-aa (changes requested):
I didn't look at the date range, because they are things to fix with the calendar first:
- We don't want to trap the focus in the calendar, unless it is inside the popover of the date picker.
- When we select a date in the calendar with the keyboard or the mouse, the focus should stay on this date.

Bonus: if you're looking to refactor, I think we could use a second class `CalendarDate`, instead of having the huge `generateViewData` method.

src/modules/base/calendar/calendar.js
```diff
@@ -171,6 +191,21 @@ export default class Calendar extends LightningElement {
         this.updateDateParameters();
     }
 
+    /**
+     * Set initial focus
+     *
+     * @type {object}
+     * @public
+     */
+    @api
```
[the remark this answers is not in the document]
lgilmet: done

src/modules/base/calendar/calendar.js
lgilmet: My reflex is always `let` :(

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -681,366 +788,494 @@ export default class InputDateRange extends LightningElement {
...
 
-        // selecting 'end date' earlier than 'start date', make that the new 'end date', void 'start date', open start date picker
-        if (this.areBothDatePresent) {
-            if (dates[0] < this._startDate) {
-                this._endDate = dates[0];
-                this._startDate = null;
+        if (keepFocusOn)
+            setTimeout(() => {
+                this.showEndDate = true;
+            }, 10);
```
nawel-aa, line 1026: Why using a timeout to set a property? It will trigger a second rerender.

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -681,366 +788,494 @@ export default class InputDateRange extends LightningElement {
...
 
     /**
-     * Sets the variable cancelBlurEndDate to false.
+     * Check for missing required data
      */
-    allowBlurEndDate(event) {
-        if (event !== undefined) {
-            this._cancelBlurEndDate = false;
-        }
+    checkInputDatesValidity() {
```
nawel-aa, line 1132: There already are public methods to check the validty of the inputs, you should use them.

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -133,6 +133,15 @@ export default class InputDateRange extends LightningElement {
...
+    savedFocus;
+
+    showEndDate = false;
+    showStartDate = false;
+    enteredStartCalendar = false;
+    enteredEndCalendar = false;
+    _selectionModeEndDate = 'interval';
 
     helpMessage;
     _valid = true;
```
nawel-aa, line 147: There are already methods and a getter (this.validity), you shouldn't need another property.

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -681,366 +788,494 @@ export default class InputDateRange extends LightningElement {
...
-                this.endDateInput.focus();
-            }
-
-            this.dispatchChange();
-        }
+    @api
+    blur() {
+        this.startDateInput.blur();
+        this.endDateInput.blur();
+        this.checkInputDatesValidity();
```
nawel-aa, line 1061: There is a method to check the validity. If the parent component wants to blur and validate, they can call `reportValidity()` after `blur()`

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -681,366 +788,494 @@ export default class InputDateRange extends LightningElement {
...
+     * Trigger focus out and validity check
+     *
+     * @public
      */
-    handlePrivateBlurEndDate(event) {
-        event.stopPropagation();
-        this.allowBlurEndDate();
-        this.handleBlurEndDate();
+    @api
+    onFocusOut() {
```
nawel-aa, line 1070: This is an event handler, it should start with "handle" and should not be decorated with `@api`, since it is not supposed to be called from the parent component.

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -681,366 +788,494 @@ export default class InputDateRange extends LightningElement {
...
+        setTimeout(() => {
+            if (!this.showStartDate && !this.showEndDate) {
+                this.interactingState.leave();
+                this.checkInputDatesValidity();
+            }
+        }, 1);
```
nawel-aa, line 1076: Why have a `setTimeout` of one and not 0?

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -681,366 +788,494 @@ export default class InputDateRange extends LightningElement {
...
-     * Handles focus for the c-calendar for end-date.
+     * Clicking either an input field or its icon to open or close its calendar
+     *
+     * @param {Event} event
      */
-    handlePrivateFocusEndDate(event) {
-        event.stopPropagation();
-        this.handleBlurEndDate();
-        this.handleFocusEndDate();
+    clickDateInput(event) {
```
nawel-aa, line 1084: By convention, this event handler should start with "handle" to be easily identified as one.

nawel-aa (changes requested):
**Calendar:** 
- When selecting a date with the keyboard, the focus should stay on the selected date.
- When tabbing in the calendar, only one date should be focusable. The date should be one of these, in this order:
  -  The last date the user has set focus on.
  - Today's date.
  - The first available date of the month.
  - The 1st of the month, if all dates are disabled.

**Input Date Range:**
- When selecting a date with the keyboard, the calendar should close and the focus be set on the calendar button icon. It should not jump to the next input. With the mouse, it should keep the current behaviour.
- Could you add to the description of the value property, that it is read-only?

src/modules/base/calendar/calendar.js
```diff
@@ -1021,6 +1071,138 @@ export default class Calendar extends LightningElement {
...
+     *
+     * @param {number} nextDate The datetime of the keyboard navigation target
+     */
+    updateFocus(nextDate) {
+        if (nextDate) {
+            let currentDate = this.date.getTime();
+
+            if (nextDate < this.min) {
+                this.date = this.min;
+                this._nextFocusDate = this.min;
```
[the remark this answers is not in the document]
lgilmet: Removed

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -157,6 +166,26 @@ export default class InputDateRange extends LightningElement {
...
+    /**
+     * Send focus date to end calendar
+     *
+     * @type {object}
+     */
+    @api
+    get focusEndDate() {
+        return this._focusEndDate;
+    }
+
```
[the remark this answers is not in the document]
lgilmet: Removed

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -397,6 +450,53 @@ export default class InputDateRange extends LightningElement {
...
+    /**
+     * End calendar.
+     *
+     * @type {element}
+     */
+    get endCalendar() {
+        return this.template.querySelector(
+            '[data-element-id="calendar-end-date"]'
+        );
+    }
```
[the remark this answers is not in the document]
lgilmet: Removed

src/modules/base/inputDateRange/inputDateRange.js
lgilmet: Refactored

src/modules/base/inputDateRange/inputDateRange.js
lgilmet: Refactored

src/modules/base/inputDateRange/inputDateRange.js
lgilmet: done

src/modules/base/inputDateRange/inputDateRange.js
lgilmet: done

src/modules/base/inputDateRange/inputDateRange.js
```diff
@@ -397,6 +450,53 @@ export default class InputDateRange extends LightningElement {
...
+
+    // UTILITIES //
```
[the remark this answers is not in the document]
lgilmet: Got it I think

PR #311, opened 2022-05-04
[fix] (abc-526): Support rich text in list item

lgilmet:
- Replaced found dead image link in storybook and examples


nawel-aa (approved):
Thanks for updating the broken images! 👍

jbverger (approved):

PR #318, opened 2022-05-10
[fix] (ABC-542): Action icons overflow the picker's limits

Felino203:
Fixed action button sizing overflow.
All actions buttons now have a different size (Medium used to be the same as large).


nawel-aa (changes requested):
Overall, it works pretty well. 👍 
Two things, though, on top of my comment in the CSS file:
- You should add a test for your new classes (if you have any questions on how to write a test, let me know).
- It would be nice if the button width wasn't growing for action icons. I know it's because the scale does not remove the space the element was taking previously, so it may prove to be difficult. But I'll let you check what you can do. :)

nawel-aa (approved):
🚀

jbverger (approved):

PR #324, opened 2022-05-10
[feat] (ABC-553): Set Dual ListBox default size to responsive

AchilleStH:
### Features
**Dual Listbox**
- Changed default size to responsive


nawel-aa (commented):
- It could be nice to add a test for the responsive size, I think we forgot when we added the option in the first place.
- The `size` attribute default value needs to be updated in [Confluence](https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/1471840257/Dual+Listbox).

But overall, everything's there! 👍

nawel-aa (approved):

jbverger (approved):

PR #327, opened 2022-05-10
[fix] (ABC-534): Calendar improve min-max use

GenevievePMD:


src/modules/base/calendar/calendar.js
```diff
@@ -761,6 +762,25 @@ export default class Calendar extends LightningElement {
         return marked;
     }
 
+    /**
+     * If invalid current day, center calendar's current day to closest date in min-max interval
+     */
+    validateCurrentDayValue() {
```
nawel-aa, line 768: - I think you may want to use `getTime()`, not `getFullYear()`? Or just compare the `Date` objects directly?
- There's also a lot of `if` and `else`, and it's hard to understand the different cases. 😄 You could refactor it using boolean variables with clear names (for example `isBeforeMin` or `isAfterMax`), and/or a variable that stores the state of the value as a string (see `handleChangeStartDate()` in the `inputDateRange` or `createInterval()` in the `timer`, for an example).
Try to first identify all the possibilities, and don't forget that the calendar can be an interval or have multiple values. 🙂

nawel-aa (changes requested):

src/modules/base/calendar/calendar.js
```diff
@@ -303,14 +312,10 @@ export default class Calendar extends LightningElement {
...
-                (x) => x.setHours(0, 0, 0, 0) !== NULL_DATE
-            );
-            this.date = this._value.length
-                ? new Date(this._value[0])
-                : DEFAULT_DATE;
-            this.updateDateParameters();
+            if (this._isConnected) {
+                this.validateCurrentDayValue();
+            }
+            this.updateDate();
```
nawel-aa, line 318: It's a detail, but since `updateDate()` is called in `validateCurrentdayValue()`, which is called in the `connectedCallback()` or just above, you probably don't need to call it here too. 🙂

nawel-aa (commented):
Nice refactor! 👌
The code looks nice, I just have a concern regarding the feature, so you'll find it on Jira. 🙂

nawel-aa (changes requested):

nawel-aa (approved):
It works! 👌

PR #328, opened 2022-05-11
[Feat] (ABC-550): Hidden Categories removed from tabs

Felino203:
### Features
Icon Picker
- Hidden categories are no longer displayed in tabs or in tabs popover


src/modules/base/iconPicker/iconPicker.js
```diff
@@ -32,7 +32,9 @@ const TABS = {
     default: 'Standard'
 };
 
-const DEFAULT_HIDDEN_CATEGORIES = ['Utility', 'Doctype', 'Action'];
+const NB_VISIBLE_TABS = 2;
+
+const DEFAULT_HIDDEN_CATEGORIES = [];
```
nawel-aa, line 37: Since it's now an empty array, you don't need to store it in a variable.

nawel-aa (changes requested):
You just forgot to remove the `@default` of the `hiddenCategories` property, but everything works! 👍

nawel-aa (approved):

PR #329, opened 2022-05-11
[Feature] (ABC-547) Add focus method icon picker

Felino203:
### Features
Icon Picker
- Added focus() public method


nawel-aa (approved):
Thank you for adding the licence too! 😃

jbverger (approved):

PR #330, opened 2022-05-11
[fix] (ABC-606) Color Gadrient drag and event propagation

AchilleStH:
### Fixes
**Color Gradient**
- The color picker indicator is now draggable
- The hue slider was set to the left when the color selected was in the corner of the gradient

### DIFF
--- a/src/modules/base/colorGradient/colorGradient.js
+++ b/src/modules/base/colorGradient/colorGradient.js
@@ -103,7 +103,7 @@ export default class ColorGradient extends LightningElement {
 
             this.setPaletteColor(this.colors.H);
             this.setSwatchColor(this.value);
-            this.setindIcatorPosition();
+            this.setIndicatorPosition();
 
             this.init = true;
         }
@@ -197,7 +197,7 @@ export default class ColorGradient extends LightningElement {
             if (this.init) {
                 this.setPaletteColor(this.colors.H);
                 this.setSwatchColor(this.value);
-                this.setindIcatorPosition();
+                this.setIndicatorPosition();
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/modules/base/colorGradient/colorGradient.js
```diff
@@ -208,6 +208,17 @@ export default class ColorGradient extends LightningElement {
...
 
+    /**
+     * Disable cursor if disabled or readOnly.
+     *
+     * @type {string}
+     */
+    get disabledClass() {
+        return this._disabled || this._readOnly
+            ? 'slds-color-picker__custom-range slds-color-picker__custom-range_disabled'
+            : 'slds-color-picker__custom-range';
```
nawel-aa, line 219: This works well, but we usually use `classSet()` for this type of application. It makes it easier to identify which classes are always present and which depend on specific conditions.

nawel-aa (changes requested):
It works, but there is a console error when the input is disabled and we try to drag the indicator, probably because you moved the mousedown handler's position. 🙂

src/modules/base/colorGradient/colorGradient.js
```diff
@@ -37,6 +37,7 @@ import {
     HSVToHSL,
     normalizeBoolean
 } from 'c/utilsPrivate';
+import { classSet } from '../utils/classSet';
```
nawel-aa, line 40: Instead of using local imports, we use the namespace `c/` to be able to change the file structure without impacting the components. So in this case, it would be `import { classSet } from 'c/utils'`

nawel-aa (approved):

jbverger (approved):

PR #333, opened 2022-05-12
[fix] (ABC-580): Input Counter - Fix '-' and '+' alignment 

Felino203:
### Fixes
Input Counter
- '-' and '+' now scale relative to top of input rather than height of element, which was bigger when a help message was shown


nawel-aa (approved):
🚀

jbverger (approved):

PR #334, opened 2022-05-12
[fix] (ABC-574): Hero Banner center text example

AchilleStH:
### Fixes
**Herro Banner**
- Changed the width of the centered example
- Renamed the centered example


nawel-aa (approved):
🚀

jbverger (approved):

PR #336, opened 2022-05-12
[feat](ABC-532) : ActivityTimeline : Add controls to change size of icons

GenevievePMD:
### Features
ActivityTimeline
- Added iconSize and itemIconSize to customize size of icons

### DIFF
--- a/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
+++ b/src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
@@ -179,7 +179,7 @@ describe('Activity Timeline', () => {
 
         return Promise.resolve().then(() => {
             const icon = element.shadowRoot.querySelector(
-                '.slds-media__figure > lightning-icon'
+                '[data-element-id="avonni-activity-timeline-icon"]'
             );
             expect(icon.iconName).toBe('standard:case');
         });
@@ -192,7 +192,7 @@ describe('Activity Timeline', () => {
 
         return Promise.resolve().then(() => {
             const icon = element.shadowRoot.querySelector(
-                '.slds-media__figure > lightning-icon'
+                '[data-element-id="avonni-activity-timeline-icon"]'
             );
             expect(icon.size).toBe('xx-small');
         });
@@ -204,7 +204,7 @@ describe('Activity Timeline', () => {
 
         return Promise.resolve().then(() => {
             const icon = element.shadowRoot.querySelector(
-                '.slds-media__figure > lightning-icon'
+                '[data-element-id="avonni-activity-timeline-icon"]'
             );
             expect(icon.size).toBe('x-small');
         });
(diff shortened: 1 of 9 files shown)


src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
```diff
@@ -185,6 +185,138 @@ describe('Activity Timeline', () => {
...
     });
 
+    // icon size
+    it('Activity Timeline: icon size (xx-small)', () => {
+        element.iconName = 'standard:case';
+        element.iconSize = 'xx-small';
+
+        return Promise.resolve().then(() => {
+            const icon = element.shadowRoot.querySelector(
+                '.slds-media__figure > lightning-icon'
```
nawel-aa, line 195: You should add a `data-element-id` attribute to the icon, so your tests don't depend on classes.

src/modules/base/activityTimeline/activityTimeline.js
```diff
@@ -180,6 +189,44 @@ export default class ActivityTimeline extends LightningElement {
         if (this._isConnected) this.initActivityTimeline();
     }
 
+    /**
+     * The size of the title's icon. Valid values are x-small, small, medium and large.
```
nawel-aa, line 193: You're missing the `xx-small` in the accepted values, for both documentation descriptions. :)

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.css
```diff
@@ -70,12 +70,68 @@
...
     width: 2px;
     position: absolute;
-    left: 2.25rem;
-    top: 5px;
+    top: 11px;
     bottom: 0;
     margin-left: 1px;
 }
 
+.avonni-activity-timeline-item-timeline-action-icon-xx-small {
```
nawel-aa, line 78: Ideally, you should use [our version of the BEM convention](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587118095/CSS#Class-naming) to name your classes, which I think in this case would be: `avonni-primitive-activity-timeline-item__action-icon_xx-small`.
It makes very long names, and I know it hasn't been used well in many components (including this one), but if you create new classes, may as well name them the way it's supposed to be. 😊

src/modules/base/primitiveActivityTimelineItem/primitiveActivityTimelineItem.js
```diff
@@ -464,6 +527,13 @@ export default class PrimitiveActivityTimelineItem extends LightningElement {
         );
     }
 
+    /**
+     * Check if the type of the icon is action
+     */
+    isActionIcon() {
```
nawel-aa, line 533: Since there is no parameter to the method and it is returning a value, this should be a `get isActionIcon()`, not a method. And you should make sure the `iconName` is a string before using `split()` on it, because `iconName` doesn't have a get/set where we transform an invalid value into a string.

nawel-aa (changes requested):
Nice catch to think of scaling the action icons! :)
Aside from my comments in the code, the main problem is that without icons (last story), the line is not aligned when we change the item icon size.

nawel-aa (approved):

jbverger (approved):

PR #337, opened 2022-05-13
[fix] (ABC-589) Rating value wrapping order

AchilleStH:
### Fixes
**Rating**
- Now displays the rating with the lower scores on top

### DIFF
--- a/src/modules/base/rating/rating.css
+++ b/src/modules/base/rating/rating.css
@@ -30,14 +30,6 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-.avonni-rating:hover .avonni-active.avonni-continuous:not(:hover) {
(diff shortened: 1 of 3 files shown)


src/modules/base/rating/rating.css
```diff
@@ -46,33 +38,52 @@
...
+.avonni-rating__container:hover .avonni-active {
+    border: 1px solid #ffffff;
+    background-color: #1b5297;
+    border-color: #1b5297;
+    color: #ffffff;
+    opacity: 1;
+}
+
+.avonni-rating__container:hover .avonni-active-star {
+    fill: #1b5297 !important;
```
nawel-aa, line 85: I'm not sure why you need the `!important`, or even the whole selector. 🤔

nawel-aa (changes requested):
The main problem is that the single selection hover is broken (see 7th story for an example).

nawel-aa (approved):
Everything seems good! 👍

PR #338, opened 2022-05-13
Fix abc 585 wizard modal small large device

ivanivanov-github:
### Fixes
**Wizard**
- Progression steps don't overflow anymore to the right in modal

### DIFF
--- a/src/modules/base/primitiveWizardNavigation/primitiveWizardNavigation.css
+++ b/src/modules/base/primitiveWizardNavigation/primitiveWizardNavigation.css
@@ -34,7 +34,7 @@
     height: 100%;
 }
 
-.avonni-primitive-wizard-navigation__style {
+.avonni-primitive-wizard-navigation__progress-step-width {
     width: 40px !important;
     min-width: 0px !important;
 }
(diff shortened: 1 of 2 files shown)


src/modules/base/primitiveWizardNavigation/primitiveWizardNavigation.css
```diff
@@ -30,6 +30,11 @@
...
  */
 
- .avonni-height_full {
+.avonni-height_full {
     height: 100%;
 }
+
+.avonni-primitive-wizard-navigation__progress-step-width {
+    width: 40px !important;
+    min-width: 0px !important;
```
nawel-aa, line 39: `!important` is almost never needed and you should avoid it as much as possible. Usually, if there is a conflict with another class, you can just make your selector more precise by chaining the parent element class, or adding two classes to the same element, for example.

In this case, you don't need this selector anyway. 🙂 The element that allows the progress indicator to overflow is the column (the `lightning-layout-item`): it grows as much as its content needs to grow, without any limit. What you want, is to make sure it grows, but up to 100% width of its parent.

nawel-aa (changes requested):

PR #340, opened 2022-05-13
[fix] (ABC- 593) Timer - broken count-down and much more

Felino203:
### Features
**Timer**
- Timer works with milliseconds.
- 2 new display formats: "mm:ss.ms" & "ss.ms"

### Fixes
**Timer**
- Countdown type now works.
- Pause now works.
- Reset method now works.
- Repeat attribute now works.
- Refactored formatting to be flexible. Can now easily add more display formats.

### Breaking Changes
**Times**
- Value attribute is now given in milliseconds instead of seconds.
- Duration caps at 24 hours.

### DIFF
--- a/src/modules/base/storybookWrappers/timer/timerButtons.js
+++ b/src/modules/base/storybookWrappers/timer/timerButtons.js
@@ -1,0 +1,65 @@
+/**
+ * BSD 3-Clause License
+ *
+ * Copyright (c) 2021, Avonni Labs, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * - Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { LightningElement, api } from 'lwc';
+
+export default class Timer extends LightningElement {
+    @api value;
+    @api variant;
+    @api type;
+    @api duration;
+    @api autoStart;
+    @api repeat;
+    @api iconName;
+    @api iconPosition;
+    @api format;
+
+    get timer() {
+        return this.template.querySelector('c-timer');
+    }
+
+    start() {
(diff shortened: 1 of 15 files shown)


src/modules/base/timer/__tests__/timer.test.js
```diff
@@ -89,28 +94,57 @@ describe('Timer', () => {
...
+        element.duration = -86200000;
+
         return Promise.resolve().then(() => {
-            expect(element.duration).toBe(DEFAULT_DURATION);
+            expect(element.duration).toBe(10000);
         });
     });
 
     // type
+    it('Timer: type = count-up', () => {
```
nawel-aa, line 118: These two tests are not really checking if the type works. What we really want when we set the type to "count-up", is that once the timer starts, the label of the lightning-button gets a bigger value over time. Here, you're testing the `start()` method and the `timerstop` event more than the type attribute. :)

src/modules/base/timer/__tests__/timer.test.js
```diff
@@ -121,19 +155,33 @@ describe('Timer', () => {
     });
 
     // repeat
-    it('Timer: repeat false', () => {
+    it('Timer: repeat = false', () => {
```
nawel-aa, line 158: Here you could also check if the button's label is up to date, since that's the end result the users will actually see.

src/modules/base/timer/timer.js
```diff
@@ -113,36 +125,36 @@ export default class Timer extends LightningElement {
         this._autoStart = normalizeBoolean(value);
 
         if (this._autoStart) {
-            this.start();
+            this.startDate = Date.now();
+            // start after short delay to account for properties initialization
+            setTimeout(() => {
+                this.start();
+            }, 10);
```
nawel-aa, line 132: You can probably set the timeout to 0, or even better, use `requestAnimationFrame(() => {})`, since what you want is to wait for the component rerender.

src/modules/base/timer/timer.js
```diff
@@ -215,21 +227,26 @@ export default class Timer extends LightningElement {
...
+     * Starting value of the timer in milliseconds. Getting this attribute will provide the current timer value in milliseconds.
      *
      * @type {number}
      * @public
      * @default 0
      */
     @api
     get value() {
-        return this._value;
+        return this._timerValue;
```
nawel-aa, line 238: Is there a specific reason to rename `_value` to `_timerValue`? It's clearer to keep the same name for the private and public properties.

src/modules/base/timer/timer.js
```diff
@@ -263,111 +280,67 @@ export default class Timer extends LightningElement {
...
      * @type {number}
      */
-    get timerValue() {
-        if (this.type === 'count-up') {
-            return this.value;
-        }
-
-        return this.duration - this.value;
+    get hours() {
+        let hours = Math.floor(this._timerValue / 60 / 60 / 1000);
```
nawel-aa, line 303: Don't forget to use `const` instead of `let`, when you can. :)

src/modules/base/timer/timer.js
```diff
@@ -566,32 +544,57 @@ export default class Timer extends LightningElement {
      * Create timer interval.
      */
     createInterval() {
```
nawel-aa, line 546: This method can probably be refactored a little to be more readable. Instead of having a lot of if/else, you could have a constant of which you change the value depending on the current state, and/or a few boolean constants with a clear name. For example:
```
const isCountUp = this.type === 'count-up';
const hasEndedCountUp = isCountUp && this._timerValue >= this._startTime + this.duration;
const hasEndedCountDown = !isCountUp && (this._timerValue <= maxTime || Math.abs(this._timerValue) >= MAX_TIMER_VALUE);

let state;
if (this.play && hasEndedCountUp) {
    state = 'RESET_COUNT_UP';
} else if (this.play && isCountUp) {
   state = 'COUNT_UP';
} else if () { ... }

switch (state) { ... }
```

nawel-aa (commented):
Works well overall! 👍
- For the stories that don't have autoplay, it would be nice to add buttons to start/stop/reset the timer. To do so, you have to create a storybook wrapper in the `storybookWrapper` folder, and change the component imported in `timer > __examples__`. Checkout the other components that use a wrapper for examples (and let me know if it's unclear).
- The value change from seconds to milliseconds makes a lot of sense, but it's a breaking change, so I would check with @jbverger that's okay, if you haven't already.
- Very small detail, but could you reorder the getters and the public methods in alphabetical order?

nawel-aa (changes requested):
Sorry I forgot to mention it before, but the examples in `__examples__` should be updated to include the new storybook wrapper. The goal is that the examples on our website match the examples in storybook. 🙂

PS: I edited your pull request comment, because we have a new "breaking change" category in [our release notes](https://www.avonnicomponents.com/release-notes).

Felino203:
Don't know if there is a way for me the test the __examples__ I added and modified. Either way, I think they should work. :)

src/modules/base/storybookWrappers/timer/timerButtons.js
```diff
@@ -44,7 +44,7 @@ export default class Timer extends LightningElement {
     @api format;
 
     get timer() {
-        return this.template.querySelector('c-timer');
```
nawel-aa: I think you may have done this change by mistake? In the documentation examples, we do want to use `avonni-timer` instead of `c-timer`: in the website, we use the base components package, and the `c` namespace is not exposed in it.
But in the storybook wrapper, you don't need to use `avonni`, you can use `c`. In any case, if you make the change in the JS file, you should rename `c-timer` to `avonni-timer` in the HTML file too, because right now the buttons don't work anymore. 🙂

nawel-aa (commented):
The storybook buttons don't work anymore (see my comment), but the component looks good to me! 👍

PR #341, opened 2022-05-13
Fix abc 566 carousel three items per panel

ivanivanov-github:
### Fixes
**Carousel**
- Displays correctly multiple items per panel
- Converts bare and border action variants to menu variant on small and large mobile devices


src/modules/base/primitiveCarouselItem/__tests__/primitiveCarouselItem.test.js
```diff
@@ -167,6 +167,44 @@ describe('Primitive Carousel Item', () => {
         });
     });
 
+    it('Primitive Carousel Item: actions variant border with large and small mobile device', () => {
```
nawel-aa, line 170: The `Object.defineProperty` is clever. 😊 But these tests will always pass, because the `lightning-button-menu` is always present in the DOM, it's just hidden by the CSS. If you want to add a test for it, you could check if the right classes are applied to both the button menu, and the regular button/button-icon.

src/modules/base/primitiveCarouselItem/primitiveCarouselItem.css
```diff
@@ -11,9 +11,28 @@
     font-weight: var(--avonni-carousel-item-title-font-weight, 600);
 }
 
+.avonni-carousel__image {
+    display: flex;
+}
+
+.avonni-carousel__button-icon-top {
```
nawel-aa, line 18: I'm not sure why you removed the surrounding margin and padding. It's good to have some space between the actions.

nawel-aa (commented):
The bug is fixed and the button menu looks good!

On top of my other comments, I noticed two things that were already there, but could be nice to fix:
- The click on an action should stop the propagation of the click event, to prevent the wrapping `<a>` tag from taking over.
- The HTML of the primitive carousel item could be split into two: one with the <a> tag, and the other without. To do so, you can use render() (see [the doc](https://lwc.dev/guide/html_templates#render-multiple-templates) or search the code base for examples).

src/modules/base/primitiveCarouselItem/__tests__/primitiveCarouselItem.test.js
```diff
@@ -200,11 +199,31 @@ describe('Primitive Carousel Item', () => {
...
             );
-            expect(action).toBeTruthy();
             expect(action.menuAlignment).toBe('auto');
         });
     });
 
+    // render html variant
+    it('Primitive Carousel Item: tag variant', () => {
+        return Promise.resolve().then(() => {
+            const tag = element.shadowRoot.querySelectorAll(
```
nawel-aa, line 209: This test pass only because you used `querySelectorAll` instead of `querySelector`, so you'll always have an empty node list, even if there is no element. 🙂 Normally, by default, the `noTag` template would be called. And if you add an href, the `tag` template would be called.

src/modules/base/primitiveCarouselItem/noTag.html
```diff
@@ -33,192 +33,6 @@
...
-                                ></lightning-menu-item>
-                            </template>
-                        </lightning-button-menu>
-                    </template>
-                </div>
-            </template>
-        </div>
-    </a>
     <div
         if:false={href}
```
nawel-aa, line 37: The `if:false` can be removed, because this template won't be used if there is an `href`.

src/modules/base/primitiveCarouselItem/noTag.html
```diff
@@ -290,6 +105,7 @@
                             menu-alignment="auto"
                             variant="border-filled"
                             data-element-id="lightning-button-menu"
+                            onclick={handleActionClick}
```
nawel-aa, line 108: If you add the `onclick` to the button menu, the `actionclick` event will be dispatched every time we click on the menu, with the wrong action name. Instead you can remove the `onclick` handlers on the menu and the items, and add an `onselect` one (see the [Salesforce doc](https://developer.salesforce.com/docs/component-library/bundle/lightning-button-menu/documentation) for a description of the event).

src/modules/base/primitiveCarouselItem/tag.html
```diff
@@ -0,0 +1,226 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+-->
+
+<template>
+    <a
+        if:true={href}
```
nawel-aa, line 37: You can remove the `if:true` since this template will never be called if there is no `href`.

nawel-aa (changes requested):
A few small changes needed.
If my comments on the tests aren't clear, don't hesitate to let me know. :)

src/modules/base/primitiveCarouselItem/primitiveCarouselItem.js
```diff
@@ -363,4 +363,13 @@ export default class PrimitiveCarouselItem extends LightningElement {
             })
         );
     }
+
+    /**
+     * Menu select event handler
+     *
+     * @param {Event}
+     */
+    handleMenuSelect(event) {
```
nawel-aa, line 372: As discussed the other day, what we want is either to use the [`select` event dispatched by the `lightning-button-menu`](https://developer.salesforce.com/docs/component-library/bundle/lightning-button-menu/documentation) (ideal), or to use the click event on each menu item (slightly less efficient).
In either case, the `actionclick` event should be dispatched only once: when we click on an action.

src/modules/base/primitiveCarouselItem/noTag.html
nawel-aa: Don't forget to make the change for all the button menus. 🙂

nawel-aa (changes requested):
- I forgot to mention it, but you may have noticed yourself: if you create several HTML templates, the CSS files that you want to attach to them need to have the same names. To prevent duplicating CSS, you can have three files: `tag.css`, `noTag.css` and `shared.css`. In `shared.css` you keep the CSS that is shared by the two templates. In the two others you add an `@import` rule (see the `primitiveSchedulerEventOccurrence` component for an example).
- Don't hesitate to click on the "resolve" button of my comments when you fix the issue. I won't be mad and it can help you see what is left to do. 😊

src/modules/base/primitiveCarouselItem/shared.css
```diff
@@ -11,9 +11,28 @@
     font-weight: var(--avonni-carousel-item-title-font-weight, 600);
 }
 
+.avonni-carousel__image {
+    display: flex;
+}
+
+.avonni-carousel__button-icon-top {
+    padding: 1px;
```
nawel-aa, line 19: My old comment from a few reviews ago still stands: why do you remove the margin/padding? And if you need to do so, I think you don't need these classes, you can change `computedLightningButtonIconActionClass` in the Javascript file, to not apply the `slds-m-...` classes. 🤔

src/modules/base/primitiveCarouselItem/shared.css
```diff
@@ -11,9 +11,28 @@
     font-weight: var(--avonni-carousel-item-title-font-weight, 600);
 }
 
+.avonni-carousel__image {
```
nawel-aa, line 14: `avonni-carousel__image` doesn't seem to be applied to anything. This declaration should be removed.

src/modules/base/primitiveCarouselItem/shared.css
```diff
@@ -11,9 +11,28 @@
...
+    padding: 1px;
+    margin: 0px;
+    min-width: 0;
+    min-height: 0;
+}
+
 .avonni-carousel__actions {
     display: flex;
     width: 100%;
+    padding: 4px;
```
nawel-aa, line 35: To make sure we always use the same kind of values for padding/margin, we try to only use the [Lightning Design System classes](https://www.lightningdesignsystem.com/utilities/padding/). You should add the one you want to `computedActionsContainerClass` in your Javascript.

src/modules/base/primitiveCarouselItem/__tests__/primitiveCarouselItem.test.js
```diff
@@ -336,4 +411,54 @@ describe('Primitive Carousel Item', () => {
...
+        element.href = 'https://www.salesforce.com';
+        element.actions = bareActions;
+        element.name = 'someName';
+
+        return Promise.resolve().then(() => {
+            const action = element.shadowRoot.querySelector(
+                '[data-element-id="lightning-button-menu"]'
+            );
+            const result = action.click();
+            expect(result).toBeFalsy();
```
nawel-aa, line 461: The way to test if an event has been prevented, is to check its [`defaultPrevented`](https://developer.mozilla.org/fr/docs/Web/API/Event/defaultPrevented) value. So instead of using `action.click()`, you could create a `CustomEvent`, make `action` dispatch it, and then check if `yourCustomEvent.defaultPrevented` is `true`. 🙂

nawel-aa (changes requested):
You still need two CSS files, with the exact same name than the HTML file (`tag.css` and `noTag.css`). Otherwise, the CSS file is not linked to the HTML. The `shared.css` file you created needs to be imported in both files through an [`@import rule`](https://developer.mozilla.org/fr/docs/Web/CSS/@import). You can check out the `primitiveSchedulerEventOccurrence` component for an example.

src/modules/base/primitiveCarouselItem/noTag.html
```diff
@@ -0,0 +1,235 @@
...
+        class="slds-carousel__panel-action"
+        data-panel-index={panelIndex}
+        data-item-index={itemIndex}
+        data-element-id="a-actions-noTag"
+        onclick={handleItemClick}
+    >
+        <div class={computedCarouselImageClass}>
+            <template if:false={isBottomPosition}>
+                <div class={computedActionsContainerClass} if:true={hasActions}>
+                    <template if:false={isMenuVariant}>
```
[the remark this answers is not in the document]
ivanivanov-github: Yes, it simplified the code quite a bit 😅.

src/modules/base/primitiveCarouselItem/shared.css
ivanivanov-github: I don't need to do so. I've removed the unnecessary classes.

nawel-aa (approved):
Everything works, congrats! 👍

jbverger (approved):

PR #345, opened 2022-05-16
[fix] (ABC-569): Color picker accessibility

AchilleStH:
### Fixes
**Color picker**
- Dropdown uses a focus trap to improve accessibility


src/modules/base/colorPicker/colorPicker.js
```diff
@@ -944,6 +949,34 @@ export default class ColorPicker extends LightningElement {
      * -------------------------------------------------------------
      */
 
+    /**
+     * Initializes the event listeners.
+     * The listeners are used to monitur a blur of the popover by clicking the menu button.
+     */
+    initEventListeners() {
+        this.template.addEventListener('mousedown', (event) => {
```
nawel-aa, line 957: I'm unsure why this block of code could not be replaced by a `onmousedown` on the `<button>` you're trying to reach.

src/modules/base/colorPicker/colorPicker.js
```diff
@@ -1133,6 +1166,16 @@ export default class ColorPicker extends LightningElement {
         if (!this.disabled) {
             this.dropdownVisible = !this.dropdownVisible;
 
+            // eslint-disable-next-line @lwc/lwc/no-async-operation
+            setTimeout(() => {
```
nawel-aa, line 1170: You can use `requestAnimationFrame()` for the same purpose, and it will make it clearer that you're waiting for a rerender.

src/modules/base/colorPicker/colorPicker.js
```diff
@@ -1133,6 +1166,16 @@ export default class ColorPicker extends LightningElement {
...
+                const tab = this.template.querySelector(
+                    '.slds-tabs_default__link'
```
nawel-aa, line 1172: You should use a [`data-element-id` attribute](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Only-data-attributes-should-be-used-in-query-selectors) instead of a class in your query selector.

nawel-aa (approved):
Everything works well! 👍
If you want to go the extra mile, you can simplify your way of closing the popover on blur. In `node_modules/lightning-base-components/src/lightning/popup`, checkout how they did it using `focusin` and `focusout`! 🙂

jbverger (approved):

PR #350, opened 2022-05-17
[Feature] (ABC-599): Created new component - chip container

Felino203:
### Features
Chip Container
- Display a list of chips from list of items (see doc)
- Possibility of displaying either an avatar or icon as media (left or right)
- Possibility of changing media size
- Collapsible and expandable

*see: https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/2147581953/Chip+Container

### DIFF
--- a/src/modules/base/chip/chip.html
+++ b/src/modules/base/chip/chip.html
@@ -40,7 +40,7 @@
                     <slot name="left"></slot>
                 </span>
             </template>
-            {label}
+            <span class="avonni-chip__chip-label"> {label} </span>
             <template if:true={showRight}>
                 <span class="slds-badge__icon slds-badge__icon_right">
                     <slot name="right"></slot>
--- a/src/modules/base/chip/chip.html
+++ b/src/modules/base/chip/chip.html
@@ -34,13 +34,20 @@
 
 <template>
     <span class={chipClass} data-element-id="span-wrapper">
-        <span class="slds-truncate">
+        <span class="slds-truncate slds-grid">
             <template if:true={showLeft}>
                 <span class="slds-badge__icon slds-badge__icon_left">
                     <slot name="left"></slot>
                 </span>
             </template>
-            <span class="avonni-chip__chip-label"> {label} </span>
+            <span
+                class="
+                    avonni-chip__chip-label
+                    slds-grid slds-grid--vertical slds-grid--align-center
+                "
+            >
+                {label}
+            </span>
             <template if:true={showRight}>
                 <span class="slds-badge__icon slds-badge__icon_right">
                     <slot name="right"></slot>
--- a/src/modules/base/chip/chip.html
+++ b/src/modules/base/chip/chip.html
@@ -43,7 +43,7 @@
             <span
                 class="
                     avonni-chip__chip-label
-                    slds-grid slds-grid--vertical slds-grid--align-center
+                    slds-grid slds-grid_vertical slds-grid_align-center
                 "
             >
                 {label}
--- a/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
@@ -5,7 +5,7 @@
  * @property {string} variant The variant changes the appearance of the chip. Accepted variants include base, brand, inverse, alt-inverse, success, info, warning, error, offline.
  * @property {boolean} outline If true, display an outline style button.
  * @property {object} avatar Avatar object. If present, the avatar is displayed to the left of the label. Overwrites icon if both medias are present.
- * @property {string} iconName Icon name. If present, the icon is displayed to the left of the label.
+ * @property {string} prefixIconName Icon name. If present, the icon is displayed to the left of the label.
  * @property {string} mediaPosition Position of the media element. Valid values include left (default) and right.
  *
  */
--- a/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
@@ -1,0 +1,20 @@
+/**
+ * @typedef {object} ChipContainerItems
+ * @name items
+ * @property {string} label Label display in the chip.
+ * @property {string} variant The variant changes the appearance of the chip. Accepted variants include base, brand, inverse, alt-inverse, success, info, warning, error, offline.
+ * @property {boolean} outline If true, display an outline style button.
+ * @property {object} avatar Avatar object. If present, the avatar is displayed to the left of the label. Overwrites icon if both medias are present.
+ * @property {string} iconName Icon name. If present, the icon is displayed to the left of the label.
+ * @property {string} mediaPosition Position of the media element. Valid values include left (default) and right.
+ *
+ */
+
+/**
+ * @typedef {object} PillContainerAvatar
+ * @name avatars
+ * @property {string} fallbackIconName The Lightning Design System name of the icon used as a fallback when the image fails to load. The initials fallback relies on this for its background color. Names are written in the format 'standard:account' where 'standard' is the category, and 'account' is the specific icon to be displayed. Only icons from the standard and custom categories are allowed.
+ * @property {string} initials If the record name contains two words, like first and last name, use the first capitalized letter of each. For records that only have a single word name, use the first two letters of that word using one capital and one lower case letter.
+ * @property {string} src Image URL.
+ * @property {string} variant The variant changes the shape of the avatar. Valid values are circle and square. Defaults to square.
+ */
--- a/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
@@ -11,7 +11,7 @@
  */
 
 /**
- * @typedef {object} PillContainerAvatar
+ * @typedef {object} ChipContainerAvatar
  * @name avatars
  * @property {string} fallbackIconName The Lightning Design System name of the icon used as a fallback when the image fails to load. The initials fallback relies on this for its background color. Names are written in the format 'standard:account' where 'standard' is the category, and 'account' is the specific icon to be displayed. Only icons from the standard and custom categories are allowed.
  * @property {string} initials If the record name contains two words, like first and last name, use the first capitalized letter of each. For records that only have a single word name, use the first two letters of that word using one capital and one lower case letter.
--- a/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
@@ -4,9 +4,9 @@
  * @property {string} label Label display in the chip.
  * @property {string} variant The variant changes the appearance of the chip. Accepted variants include base, brand, inverse, alt-inverse, success, info, warning, error, offline.
  * @property {boolean} outline If true, display an outline style button.
- * @property {object} avatar Avatar object. If present, the avatar is displayed to the left of the label. Overwrites icon if both medias are present.
- * @property {string} prefixIconName Icon name. If present, the icon is displayed to the left of the label.
- * @property {string} mediaPosition Position of the media element. Valid values include left (default) and right.
+ * @property {object} avatar Avatar object. If present, the avatar is displayed to the left of the label.
+ * @property {string} prefixIconName Prefix Icon name. If present, the icon is displayed to the left of the label.
+ * @property {string} suffixIconName Suffix Icon name. If present, the icon is displayed to the right of the label.
  *
  */
 
@@ -17,4 +17,5 @@
  * @property {string} initials If the record name contains two words, like first and last name, use the first capitalized letter of each. For records that only have a single word name, use the first two letters of that word using one capital and one lower case letter.
  * @property {string} src Image URL.
  * @property {string} variant The variant changes the shape of the avatar. Valid values are circle and square. Defaults to square.
+ * @property {string} position Position of the avatar. Valid values include left (default) and right.
  */
--- a/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
@@ -17,5 +17,5 @@
  * @property {string} initials If the record name contains two words, like first and last name, use the first capitalized letter of each. For records that only have a single word name, use the first two letters of that word using one capital and one lower case letter.
  * @property {string} src Image URL.
  * @property {string} variant The variant changes the shape of the avatar. Valid values are circle and square. Defaults to square.
- * @property {string} position Position of the avatar. Valid values include left (default) and right.
+ * @property {string} position Position of the avatar relative to the text. Valid values include left and right. Defaults to left.
  */
--- a/src/modules/base/chipContainer/__docs__/chipContainer.stories.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.stories.js
@@ -54,7 +54,7 @@ export default {
             description:
                 'The alternative text used to describe the chip container.',
             table: {
-                defaultValue: { summary: 'Task List' },
+                defaultValue: { summary: 'Selected Options:' },
                 type: { summary: 'string' }
             }
         },
--- a/src/modules/base/chipContainer/__docs__/chipContainer.stories.js
+++ b/src/modules/base/chipContainer/__docs__/chipContainer.stories.js
@@ -1,0 +1,57 @@
+/**
+ * BSD 3-Clause License
+ *
+ * Copyright (c) 2021, Avonni Labs, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * - Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * - Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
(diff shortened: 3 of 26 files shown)


src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,257 @@
...
+            resizeObserver.observe(
+                this.template.querySelector(
+                    '[data-element-id="chip-container-list"]'
+                )
+            );
+            this._resizeObserver = resizeObserver;
+        }
+    }
+
+    calculateWrappedNodes() {
```
nawel-aa, line 230: It's just a suggestion, but I think it would be cleaner to add a `hidden` attribute to your primitive chip and let it handle the class change.
- You would have to change your HTML so the chip **is** the list item (in that case use [aria roles](https://developer.mozilla.org/fr/docs/Web/Accessibility/ARIA/Roles/listbox_role) instead of the `<ul>` and `<li>` tags).
- You would update the item objects when the `is-expanded` / `is-collapsible` attributes are changed, instead of doing the calculation at every rerender. Ideally, to not change the value given by the user, you should create a `computedItems` to which you would add the key you need (see the `list` for an example).

Another (and simpler) way to go, is to not change anything to the chips, and simply rely on the fixed height and `overflow: hidden` of the parent container. That's what the lazy pill container is doing. 😊

src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,257 @@
...
+
+    computeWrapperStyle() {
+        const tallest = [
+            ...this.template.querySelectorAll(
+                '[data-element-id="chip-container-list"] *'
+            )
+        ].sort((prev, next) => next.clientHeight - prev.clientHeight)[0];
+        this.template.querySelector(
+            '[data-element-id="div-wrapper"]'
+        ).style.height = `${tallest.clientHeight}px`;
```
nawel-aa, line 255: This will raise an error if no items were set, since `tallest` will be undefined.

src/modules/base/chipContainer/chipContainer.css
```diff
@@ -0,0 +1,56 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+.slds-listbox_selection-group {
+    height: fit-content;
+}
+
+.avonni-chip-container__wrapper_is-collapsed {
```
nawel-aa, line 37: You should remove unused selectors.

src/modules/base/chipContainer/chipContainer.css
```diff
@@ -0,0 +1,56 @@
...
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+.slds-listbox_selection-group {
```
nawel-aa, line 33: If you need to overwrite an slds class, you should either not use it, or you should apply a second custom class to make sure your CSS selector is specific to this component. Otherwise, once in Avonni, the CSS won't be split into components, and this rule may apply to other components.

src/modules/base/chipContainer/chipContainer.css
```diff
@@ -0,0 +1,56 @@
...
+.avonni-chip-container__wrapper_is-collapsed {
+    /* max-width: 400px; */
+}
+
+.avonni-chip-container__wrapper_is-collapsed .wrapped {
+    visibility: hidden;
+    height: 0px;
+}
+
+.avonni-chip-container__wrapper_small {
```
nawel-aa, line 46: I don't think these classes are used.

src/modules/base/chipContainer/__tests__/chipContainer.test.js
```diff
@@ -0,0 +1,186 @@
...
+    });
+
+    it('Default attributes', () => {
+        expect(element.items).toEqual([]);
+        expect(element.alternativeText).toEqual(DEFAULT_ALTERNATIVE_TEXT);
+        expect(element.isCollapsible).toEqual(false);
+        expect(element.isExpanded).toEqual(false);
+    });
+
+    /* ----- ATTRIBUTES ----- */
```
nawel-aa, line 88: We usually try to write the attributes' tests in alphabetical order (`alternative-text`, then `is-collapsible`, etc.). It's mostly important for bigger test files, but always a good habit to have. 😊

src/modules/base/chipContainer/__tests__/chipContainer.test.js
```diff
@@ -0,0 +1,186 @@
...
+
+    // items
+    it('items', () => {
+        element.items = MOCK_ITEM_SET;
+        return Promise.resolve().then(() => {
+            const itemsContainer = element.shadowRoot.querySelector(
+                '[data-element-id="chip-container-list"]'
+            );
+            expect(itemsContainer.childElementCount).toBe(3);
```
nawel-aa, line 97: You can also test if your items did receive all of their properties. You can check out the pill container items' test for an example.

src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,257 @@
...
+ */
+
+import { LightningElement, api } from 'lwc';
+import { generateUUID } from 'c/inputUtils';
+import { normalizeString, normalizeBoolean } from 'c/utilsPrivate';
+import { classSet } from 'c/utils';
+import { AvonniResizeObserver } from 'c/resizeObserver';
+
+const DEFAULT_ALTERNATIVE_TEXT = 'Task list';
+
```
nawel-aa, line 40: We need the [@class JSDoc](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#Allowed-block-tags-for-classes) comment for this component to appear in our doc, with an `@public` tag.

src/modules/base/primitiveChip/primitiveChip.js
```diff
@@ -0,0 +1,212 @@
...
+     *  @type {Object}
+     *  @default null
+     */
+    @api
+    get avatar() {
+        return this._avatar;
+    }
+    set avatar(value) {
+        if (value) {
+            const tempAvatar = JSON.parse(JSON.stringify(value));
```
nawel-aa, line 139: You can use `normalizeObject(value)` to make sure your `JSON.parse(JSON.stringify())` doesn't raise an error if the user gave the wrong value format. I also personally prefer using the `deepCopy` utility, I find that the purpose is clearer, but that's your choice. 😊

src/modules/base/primitiveChip/__tests__/primitiveChip.test.js
```diff
@@ -0,0 +1,196 @@
...
+        expect(element.label).toEqual('');
+        expect(element.outline).toEqual(false);
+        expect(element.avatar).toEqual(null);
+        expect(element.variant).toEqual('base');
+        expect(element.iconName).toEqual(DEFAULT_ICON_NAME);
+        expect(element.iconSize).toEqual('x-small');
+        expect(element.mediaPosition).toEqual('left');
+    });
+
+    /* ----- ATTRIBUTES ----- */
```
nawel-aa, line 76: You miss a test for `iconName`.

nawel-aa (changes requested):
- You should use a storybook wrapper to add a `max-width` around your component, so the difference between the collapsed and non-collapsed container is obvious in your stories. You can check the pill container's storybook wrapper for an example.
- You should add [an example per story](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#Documenting-examples) in the `__examples__` folder.
- The avatars and items objects that you documented as tables in Confluence [should also be documented](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#Documenting-objects) in a `.jsdoc.js` file, in the `__doc__` folder.

src/modules/base/chipContainer/chipContainer.js
Felino203: Since I was asked to remove the option to resize chips, all chips should be of similar enough size to use the "lazy" method. The reason I had to hide them, was because some large chips where visible when wrapped even because the wrapper also scaled with their size. I will remove the hidden class now that it is not needed.

Felino203: Nevermind... there is too much of a height difference between an chip with an avatar, and one with only a label. I will have to implement the longer method.

src/modules/base/primitiveChip/__tests__/primitiveChip.test.js
```diff
@@ -0,0 +1,179 @@
...
+        });
+    });
+
+    // iconName
+    it('iconName', () => {
+        element.iconName = 'utility:user';
+
+        return Promise.resolve().then(() => {
+            expect(element.iconName).toBe('utility:user');
+            // cannot test the value of the html element that displays the icon
```
nawel-aa, line 109: Hmmm... Yes you can. If you retrieve the left icon element and use `expect(icon.iconName).toBe('utility:user')`, it should work.

src/modules/base/primitiveChip/primitiveChip.css
```diff
@@ -0,0 +1,4 @@
+.avonni-primitive-chip__hidden {
```
nawel-aa, line 1: It would make more sense to use `display: none` instead of `visibility`, since you don't want the element to take any room. And you can even use the class [`slds-hide`](https://www.lightningdesignsystem.com/utilities/visibility/#Hide-Show) instead of using a custom CSS file. 🙂

nawel-aa (changes requested):
- You still miss [an example per story](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#Documenting-examples) in the __examples__ folder. You need to create one component per storybook example, with the exact same settings as used in the story. These components will be copied and displayed in [our website](https://www.avonnicomponents.com/) when we do the package release.
- The tests are broken because of the last changes.

src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,238 @@
...
+    _isCollapsible = false;
+    _isExpanded = false;
+
+    _wrappedChips = 0;
+    _resizeObserver;
+
+    renderedCallback() {
+        if (this.showMore) {
+            // set the wrapper height to be as high as the biggest element
+            this.calculateWrappedNodes();
```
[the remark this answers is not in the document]
Felino203: It is needed here because on initialization, the resize observer callback is not called. So we have to manually call it once on render.

nawel-aa (approved):
🎉

jbverger (approved):

src/modules/base/chipContainer/__docs__/chipContainer.jsdoc.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @typedef {object} ChipContainerItems
```
nawel-aa, line 2: Don't forget to update this doc now that `mediaPosition` doesn't exist anymore (see the changes I made in Confluence). 🙂

src/modules/base/chipContainer/__docs__/chipContainer.stories.js
```diff
@@ -0,0 +1,108 @@
...
+
+const Template = (args) => ChipContainer(args);
+const TemplateWithMaxWidth = (args) => MaxWidthChipContainer(args);
+
+export const Base = Template.bind({});
+Base.args = {
+    items: ITEMS,
+    alternativeText: DEFAULT_ALTERNATIVE_TEXT,
+    isCollapsible: false,
+    isExpanded: false
```
nawel-aa, line 99: Not a big deal, but it may save you time on other components: since these are the default values of `alternativeText`, `isCollapsible` and `isExpanded`, you don't need to add them to the story. 😊

src/modules/base/chipContainer/__examples__/base/base.js
```diff
@@ -0,0 +1,67 @@
...
+            }
+        },
+        {
+            label: 'Fifth chip',
+            outline: false
+        },
+        {
+            label: 'Sixth chip',
+            outline: true,
+            prefixIconName: 'utility:down', // not shown since there is an avatar
```
nawel-aa, line 47: Don't forget to remove this comment when you make the change asked for in Jira. 🙂

src/modules/base/chipContainer/__examples__/collapsed/collapsed.js
```diff
@@ -0,0 +1,67 @@
...
+            }
+        },
+        {
+            label: 'Fifth chip',
+            outline: false
+        },
+        {
+            label: 'Sixth chip',
+            outline: true,
+            prefixIconName: 'utility:down', // not shown since there is an avatar
```
nawel-aa, line 47: Don't forget to remove this comment when you make the change asked for in Jira. 😊

src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,263 @@
...
+        }
+        return `+${this._wrappedChips} more`;
+    }
+
+    /**
+     * Computes the wrapper class. Adds a class if showMore button is visible.
+     */
+    get computedWrapperClass() {
+        return classSet('slds-listbox_selection-group').add({
+            'avonni-chip-container__list-box_height': true,
```
nawel-aa, line 181: If the class is always present, you can chain it inside the `classSet()` separated by a space. So in your case: `classSet('slds-listbox_selection-group avonni-chip-container__list-box_height')`.

src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * True if the "show more" button should be visible.
+     *
+     * @type {boolean}
+     */
+    get showMore() {
+        let isCollapsible =
+            (!this.isCollapsible && !this.isExpanded) || this.isCollapsible;
+        return isCollapsible && !this.computedIsExpanded;
```
nawel-aa, line 161: I'm not sure of it because I've twisted my mind trying to understand this complicated boolean chain. 😅 But it feels like you just need to return `this.isCollapsible && !this.isExpanded`?

src/modules/base/chipContainer/chipContainer.js
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * Get a unique key for li element.
+     */
+    get uniqueKey() {
+        return generateUUID();
+    }
+
+    set hasMedia(value) {
```
nawel-aa, line 193: Since you only set `hasMedia` with a `true` value, I think for more clarity, you could just add this block of code to `checkForMedia()`? That way you don't even need a `_hasMedia` property.

src/modules/base/primitiveChip/primitiveChip.js
```diff
@@ -0,0 +1,235 @@
...
+     *  @public
+     *  @type {Object}
+     *  @default null
+     */
+    @api
+    get avatar() {
+        return this._avatar;
+    }
+    set avatar(value) {
+        if (value) {
```
nawel-aa, line 135: What if there used to be an avatar, but now I want to remove it? 😊 It's unlikely in this case because it's a primitive you have control over, but as a general rule, you should allow to remove a value already set.

src/modules/base/primitiveChip/primitiveChip.js
```diff
@@ -0,0 +1,235 @@
...
+     *  @type {string}
+     *  @default x-small
+     */
+    @api
+    get prefixIconName() {
+        return this._prefixIconName;
+    }
+    set prefixIconName(value) {
+        this._prefixIconName = value;
+        this._showIcon = true;
```
nawel-aa, line 155: What if the `prefixIconName` was set to `undefined` or `null`? `this._showIcon` would still be set to `true`. You can probably check if there is a prefix or suffix icon in the `get showIcon()` directly. 🙂

nawel-aa (changes requested):
It still works, and the consistent height looks much better!
See also my comment on Jira.

nawel-aa (commented):
Only little refactoring comments. Everything works well. 🙂

src/modules/base/chip/chip.html
```diff
@@ -34,13 +34,20 @@
...
             <template if:true={showLeft}>
                 <span class="slds-badge__icon slds-badge__icon_left">
                     <slot name="left"></slot>
                 </span>
             </template>
-            {label}
+            <span
+                class="
+                    avonni-chip__chip-label
+                    slds-grid slds-grid--vertical slds-grid--align-center
```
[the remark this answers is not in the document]
Felino203: They seem to exist since they have an effect on the component style. I downloaded a vscode LWC snippets extension, and these were part of the LWC library. I guess LWC changed their naming conventions at one point and these are either new classes, or old ones. I changed them to the new conventions, but I cant make it work with slds-grid_vertical-align-center when there is no media (I want justify center on vertical flex, not align center on horizontal flex).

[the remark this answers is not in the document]
nawel-aa: Oh, I see, they do seem to have changed the dashes for underscores. I see you found the new syntax for the classes you actually needed. Nice! 🙂

nawel-aa (approved):
Looks good to go! 🚀

jbverger (approved):

PR #351, opened 2022-05-17
[fix] (ABC-607): Button menu: Fixed auto focus on menu item when opening menu

GenevievePMD:
### Fixes
Button menu
- Fixed auto-focus on first menu item when opening menu
- Added keyboard navigation (arrow up and down) to change focus


src/modules/base/buttonMenu/buttonMenu.js
```diff
@@ -772,6 +768,54 @@ export default class ButtonMenu extends LightningElement {
...
+                        : menuItemIndex + 1;
+
+                if (nextIndex >= this.getMenuItems().length) nextIndex = 0;
+                else if (nextIndex < 0)
+                    nextIndex = this.getMenuItems().length - 1;
+                this.focusOnMenuItem(nextIndex);
+                break;
+            }
+            case keyCodes.escape: {
+                if (this._dropdownVisibles) {
```
nawel-aa, line 807: `_dropdownVisibles` does not exist so this line won't work. Did you mean `_dropdownVisible`?
Also, as per the [brand new rule](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Avoid-using-one-liners), you should wrap your one line if/else in brackets. 🙂

nawel-aa (changes requested):
Everything work, except pressing "Escape", due to a little typo. 🙂

nawel-aa (approved):
👍

PR #352, opened 2022-05-18
[feat](ABC-531) Activity timeline: Added option to specify datetime format

GenevievePMD:
### Features
Activity Timeline
- Added option to specify a datetime format


nawel-aa (changes requested):
Only one tiny thing! 😊
I think the date format should not be a way to hide the date. If the date format is invalid, the default format should be used, the date should not disappear.

src/modules/base/activityTimeline/activityTimeline.js
```diff
@@ -278,7 +278,33 @@ export default class ActivityTimeline extends LightningElement {
...
+    /**
+     * If true, the date of each item is hidden.
+     *
+     * @public
+     * @type {boolean}
+     * @default false
+     */
+    @api
+    get hideItemDate() {
+        return this._closed;
```
nawel-aa, line 300: I suppose it was supposed to be `this._hideItemDate`, instead of `this._closed`?

src/modules/base/activityTimeline/activityTimeline.js
```diff
@@ -278,7 +278,33 @@ export default class ActivityTimeline extends LightningElement {
         return this._itemDateFormat;
     }
     set itemDateFormat(value) {
-        this._itemDateFormat = value && typeof value === 'string' ? value : '';
+        if (this._hideItemDate) {
+            this._itemDateFormat = '';
```
nawel-aa, line 282: What if `hideDateFormat` is set to `false` later? How do we get the format back? 😊  It would be safer to either:
- Pass a `computedDateFormat` (or other name of your choice) as the primitives `date-format`. It would return an empty string if `hideItemFormat` is true, or the `itemDateFormat` otherwise.
- Pass `hideItemDate` to the primitives, and use it in their `formattedDate`.

src/modules/base/activityTimeline/__examples__/withoutIcons/withoutIcons.html
```diff
@@ -4,6 +4,7 @@
         actions={actions}
         title="Activity Timeline"
         icon-name="standard:timesheet_entry"
+        hide-item-date="true"
```
nawel-aa, line 7: Since it would be `true` whatever the string was (it would still be `true` if you had written `hide-item-date="false"`), we just write the name of the attribute (`hide-item-date`), without a value.

nawel-aa (changes requested):
I'm sorry I hadn't read the Jira task instructions last time, where it specifically say to hide the date when the format is empty. 😅 I still think it's clearer this way, but @jbverger's opinion may differ.

nawel-aa (approved):
Looks perfect to me! 👍

jbverger (approved):

PR #353, opened 2022-05-20
Bugfix (ABC-621): reworked the range component

Felino203:
### Features
Range
- Range is now composed of 2 inputs overlayed

### Fixes
Range
- New structure removes the "wobble" effect on range change
- If both inputs have the same value, input thumbs overlap

### DIFF
--- a/src/modules/base/range/__tests__/range.test.js
+++ b/src/modules/base/range/__tests__/range.test.js
@@ -96,7 +96,7 @@ describe('Range', () => {
 
         return Promise.resolve().then(() => {
             const inputs = element.shadowRoot.querySelectorAll(
-                '.slds-slider__range'
+                '[data-group-name="input"]'
             );
             inputs.forEach((input) => {
                 expect(input.disabled).toBeFalsy();
@@ -108,9 +108,16 @@ describe('Range', () => {
         element.disabled = true;
 
         return Promise.resolve().then(() => {
-            const inputs = element.shadowRoot.querySelectorAll('.slds');
+            const inputs = element.shadowRoot.querySelectorAll(
+                '[data-group-name="input"]'
+            );
             inputs.forEach((input) => {
                 expect(input.disabled).toBeTruthy();
+                expect(
+                    element.shadowRoot
+                        .querySelector('[data-element-id="progress-bar"]')
+                        .classList.contains('avonni-range__progress_disabled')
+                ).toBe(true);
             });
         });
     });
@@ -254,7 +261,7 @@ describe('Range', () => {
 
         return Promise.resolve().then(() => {
             const inputs = element.shadowRoot.querySelectorAll(
-                '.slds-slider__range'
+                '[data-group-name="input"]'
             );
             inputs.forEach((input) => {
                 expect(input.step).toBe('3');
@@ -338,7 +345,7 @@ describe('Range', () => {
                 '[data-element-id="horizontal-unit-container"]'
             );
             const bubbles = element.shadowRoot.querySelectorAll(
-                '.avonni-range__bubble'
+                '[data-group-name="bubble"]'
             );
 
             expect(wrapper).toBeFalsy();
--- a/src/modules/base/range/__tests__/range.test.js
+++ b/src/modules/base/range/__tests__/range.test.js
@@ -36,6 +36,7 @@ import { FieldConstraintApiWithProxyInput } from 'c/inputUtils';
 import Range from 'c/range';
 
 // Not tested because not used:
+// Debounce on mousemove of 50ms
 // messageWhenRangeOverflow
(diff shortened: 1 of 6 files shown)


src/modules/base/range/range.css
```diff
@@ -174,3 +174,41 @@
     font-style: var(--avonni-range-header-font-style, normal);
     font-weight: var(--avonni-range-header-font-weight, 400);
 }
+
+.slds-slider__range {
```
nawel-aa, line 178: There is already a `.slds-slider__range` selector at the top of the file. You should add your new properties inside it.
It's also better to not change the SLDS classes directly. But this component was already doing it, so I won't blame you for that. 😊

src/modules/base/range/range.js
```diff
@@ -148,10 +148,34 @@ export default class Range extends LightningElement {
...
 
     renderedCallback() {
+        this._leftInput = this.template.querySelector(
+            '.avonni-range__slider-left'
+        );
+        this._rightInput = this.template.querySelector(
+            '.avonni-range__slider-right'
+        );
+        this._progress = this.template.querySelector('.avonni-range__progress');
+        this.template.addEventListener('mousemove', (event) => {
```
nawel-aa, line 166: Since `this.template` is not recreated at every render, we usually add the event listeners in the `connectedCallback()` (according to [Salesforce doc](https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.events_handling) you should even add it in the constructor).
I'm also suspecting that adding it in the `renderedCallback()` could lead to a memory leak.

src/modules/base/range/range.js
```diff
@@ -148,10 +148,34 @@ export default class Range extends LightningElement {
...
     _helpMessage;
+    _leftInput;
+    _rightInput;
+    _progress;
+    _moveEventWait = false;
 
     _rendered = false;
 
     renderedCallback() {
+        this._leftInput = this.template.querySelector(
```
nawel-aa, line 159: - If you're not going to change the value of `_leftInput`, you can use a `get leftInput()` instead (same for `_rightInput` and `_progress`).
- [You should use a `data-element-id` attribute](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Only-data-attributes-should-be-used-in-query-selectors) instead of relying on a CSS class in you query selectors..

src/modules/base/range/range.js
```diff
@@ -601,35 +599,86 @@ export default class Range extends LightningElement {
...
+     * @param {Event} event
+     */
+    updateInputRange(event) {
+        let minVal = parseInt(this._leftInput.value, 10);
+        let maxVal = parseInt(this._rightInput.value, 10);
+        if (maxVal - minVal >= 0 && maxVal <= this._rightInput.max) {
+            this.updateMinProgressBar(minVal);
+            this.updateMaxProgressBar(maxVal);
+        } else if (maxVal - minVal < 0) {
+            if (event.target.classList.contains('avonni-range__slider-left')) {
```
nawel-aa, line 652: If you can, you should use a data attribute to identify which slider it is. For example in this case:
`event.dataset.elementId === 'input-left'`

nawel-aa (changes requested):

src/modules/base/range/__tests__/range.test.js
```diff
@@ -108,9 +108,7 @@ describe('Range', () => {
         element.disabled = true;
 
         return Promise.resolve().then(() => {
-            const inputs = element.shadowRoot.querySelectorAll(
-                '[data-element-id="input"]'
-            );
+            const inputs = element.shadowRoot.querySelectorAll('.slds');
```
nawel-aa, line 111: `inputs` will always be empty because no element has the class `slds`. If you want both inputs, you can either:
- Use the technique that was used before, with `[data-element-id^="input"]` as a query selector. 
- Add a [`data-group-name`](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Only-data-attributes-should-be-used-in-query-selectors) to both inputs.

src/modules/base/range/__tests__/range.test.js
```diff
@@ -256,7 +254,7 @@ describe('Range', () => {
 
         return Promise.resolve().then(() => {
             const inputs = element.shadowRoot.querySelectorAll(
-                '[data-element-id="input"]'
+                '.slds-slider__range'
```
nawel-aa, line 257: Even in the tests, ideally the query selectors should not rely on classes. See previous comment for the two options you have to get both inputs.

nawel-aa (changes requested):
Very small changes in the tests, the feature itself still works like a charm. 😊

PR #358, opened 2022-05-24
[feat] (ABC-617): Kanban component

AchilleStH:
### Features
**Kanban**
- New kanban component created


src/modules/base/kanban/kanban.js
```diff
@@ -0,0 +1,1281 @@
...
+    animateTiles(groups) {
+        const fields = this.template.querySelectorAll(
+            '[data-element-id="avonni-kanban__field"]'
+        );
+        const container = this.template.querySelector(
+            '[data-element-id="avonni-kanban__field_container"]'
+        );
+
+        fields.forEach((field, i) => {
+            if (this.variant === 'base') field.style.height = 'fit-content';
```
nawel-aa, line 398: Per [the new rule](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Avoid-using-one-liners) requested by JB, your ifs should be wrapped in curly braces. 🙂

src/modules/base/kanban/kanban.js
```diff
@@ -0,0 +1,1281 @@
...
+            if (field.offsetHeight >= container.offsetHeight)
+                field.style.height = `${container.offsetHeight}px`;
```
nawel-aa, line 400: I'm not sure why you need to set the height of the fields, so I may be wrong. 😊 But if it is for the visual boxes to take only the height of their content, you can just add a `<div>` inside the one that is in flex, and apply the style to it, instead of applying it to the "column" `<div>`.

src/modules/base/kanban/kanban.js
```diff
@@ -0,0 +1,1281 @@
...
+        this.resetAnimations(groups);
+    }
+
+    /**
+     *
+     * Updates the summary value gradually
+     * @param {object} group Group containing the summary value to animate
+     */
+    animateSummary(group) {
+        const SUMMARY_UPDATE_SPEED = 300;
```
nawel-aa, line 428: Nice idea to make `SUMMARY_UPDATE_SPEED` a constant! You should declare it at the top of the file, so it would be an actual global constant, and to make it easier to find it if we want to change the speed.

src/modules/base/kanban/kanban.js
```diff
@@ -0,0 +1,1281 @@
...
+            window.clearInterval(this._scrollingInterval);
+            this._scrollingInterval = null;
+        }
+    }
+
+    /**
+     *
+     * Limits the height of the fields to prevent overflow
+     */
+    capFieldHeight() {
```
nawel-aa, line 508: Same thing as above: I haven't dived into it, so correct me if I'm wrong. But it feels like if the parent element has a `max-height` of 100%, and all of its children too, you shouldn't need to compute a height limit in your Javascript, everything should be handled with CSS. 🤔

src/modules/base/kanban/kanban.js
```diff
@@ -0,0 +1,1281 @@
...
+        this._droppedTileHeight = 0;
+    }
+
+    /**
+     * Calculates the boundaries of the kanban to prevent from dragging outside of the container
+     * @param {EventTarget} currentTarget
+     *
+     */
+    computeKanbanBoundaries(currentTarget) {
+        this._kanbanPos.top = currentTarget.getBoundingClientRect().top;
```
nawel-aa, line 551: Can't you use `currentTarget.getBoundingClientRect()` for all the sides (bottom and right included)?

src/modules/base/kanban/kanban.js
```diff
@@ -0,0 +1,1281 @@
...
+        this.dispatchEvent(
+            new CustomEvent('change', {
+                detail: {
+                    id: arr[fromIndex].id,
+                    action: arr[fromIndex][this._groupFieldName],
+                    records: this.records
+                },
+                composed: false,
+                bubbles: true,
+                cancelable: true
```
nawel-aa, line 1201: If the event is cancelable, it should actually cancel the change when `preventDefault()` is called on it by the parent component.
I don't know if it is what we want (maybe discuss it with @jbverger), but if it is, you should implement it. 😊 You can see an example of implementation in the `dispatchClick()` of the `primitiveTreeItem`.

nawel-aa (changes requested):
Really nice job! 👌  
I didn't review the tests, because you'll have enough changes to make that they may change. I also have two more comments:
- Since you have an extra "Types and Type Attributes" table in [the Confluence doc](https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/366444586/Kanban#Types-and-Type-Attributes), you need to add a `.md` file to reflect it in [our documentation website](https://www.avonnicomponents.com/). You can check out `scheduler/__docs__/scheduler.md` or `datatable/__docs__/datatable.md `for examples.
- Only a left click should trigger the drag and drop behaviour.

src/modules/base/kanban/kanban.css
```diff
@@ -0,0 +1,421 @@
...
+
+.avonni-kanban__path_item {
+    min-width: var(--avonni-kanban-column-sizing-min-width, 14em);
+    max-width: calc(
+        var(--avonni-kanban-column-sizing-max-width, 22.5em) - 0.5rem
+    );
+    background-color: var(
+        --avonni-kanban-path-color-background,
+        #014486
+    ) !important;
```
nawel-aa, line 23: If you can avoid using `!important`, it's better. I'm assuming an SLDS class is bothering you. If so, make sure you can't recreate it instead of using it.

src/modules/base/kanban/kanban.css
```diff
@@ -0,0 +1,421 @@
...
+    height: 100%;
+    max-height: 100%;
+}
+
+.avonni-kanban__field_container {
+    max-height: calc(100% - 1rem);
+    height: 100%;
+    right: 0;
+    left: 0;
+    display: flex;
```
nawel-aa, line 51: Instead of using `display: flex` everywhere, you can usually use `slds-grid`. Checkout [the LDS utilities](https://www.lightningdesignsystem.com/utilities/grid/), almost all the flex properties have a corresponding class.

src/modules/base/kanban/kanban.css
```diff
@@ -0,0 +1,421 @@
...
+    height: calc(100% - 3rem);
+}
+
+.avonni-kanban__variant_path .avonni-kanban__field {
+    height: calc(100% - 1rem);
+    display: flex;
+    flex-direction: column;
+}
+
+.avonni-kanban__tiles_container {
```
nawel-aa, line 65: Can you use the [scrollable class](https://www.lightningdesignsystem.com/utilities/scrollable/) utilities instead?

src/modules/base/kanban/kanban.css
```diff
@@ -0,0 +1,421 @@
...
+.avonni-kanban__variant_base .avonni-kanban__tile {
+    border-bottom: 1px solid #c0c0c0;
+}
+
+.avonni-kanban__tile-title {
+    font-weight: 500;
+}
+
+.avonni-kanban__variant_path .avonni-kanban__tile-title {
+    display: none;
```
nawel-aa, line 167: Can you use `slds-hide`, instead?

src/modules/base/kanban/kanban.css
```diff
@@ -0,0 +1,421 @@
...
+}
+
+.avonni-kanban_expandable_container {
+    overflow-y: auto;
+    height: 100%;
+}
+
+.avonni-kanban__expandable_section .avonni-kanban__tiles_container {
+    height: 100%;
+    /* position: sticky;
```
nawel-aa, line 366: Don't forget to remove comments.

src/modules/base/kanban/kanban.css
```diff
@@ -0,0 +1,421 @@
...
+.avonni-kanban__variant_path .avonni-kanban__expandable_section_header:hover {
+    background-color: #fafafa;
+}
+
+.avonni-kanban__expandable_section_collapsed .avonni-kanban__chevron {
+    transform: rotate(-90deg);
+}
+
+.avonni-kanban__expandable_section_collapsed .avonni-kanban__field_container {
+    display: none;
```
nawel-aa, line 420: Can you use `slds-hide`, instead?

src/modules/base/kanban/__docs__/kanban.jsdoc.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @memberof stylingHooks
+ * @name --avonni-kanban-summary-color-text
+ * @default #10AA52
+ * @type color
+ */
+/**
+ * @memberof stylingHooks
+ * @name --avonni-kanban-summarize-font-size
+ * @default 1.25rem
```
nawel-aa, line 73: That's not the default value used in the CSS for the font-size.

src/modules/base/kanban/__docs__/kanban.jsdoc.js
```diff
@@ -0,0 +1,81 @@
...
+ * @property {object} typeAttributes Provides custom formatting with component attributes for the data type. For example, currency-code for the currency type.
+ */
+
+/**
+ * @namespace stylingHooks
+ */
+/**
+ * @memberof stylingHooks
+ * @name --avonni-kanban-column-sizing-min-width
+ * @default 14.5em
```
nawel-aa, line 36: That's not the default value used for the min-width in the CSS.

src/modules/base/kanban/__tests__/kanban.test.js
```diff
@@ -0,0 +1,251 @@
...
+    });
+
+    beforeEach(() => {
+        element = createElement('avonni-kanban', {
+            is: Kanban
+        });
+        document.body.appendChild(element);
+    });
+
+    it('Kanban : Default attributes', () => {
```
nawel-aa, line 53: `hideHeader` and `subGroupFieldName` should appear in the default attributes.

src/modules/base/kanban/__tests__/kanban.test.js
```diff
@@ -0,0 +1,251 @@
...
+        return Promise.resolve().then(() => {
+            const tile = element.shadowRoot.querySelector(
+                '[data-element-id="avonni-kanban__tile"]'
+            );
+            tile.dispatchEvent(new MouseEvent('mousedown'));
+            expect(tile.classList).not.toContain('avonni-kanban__dragged');
+        });
+    });
+
+    it('Kanban : tile drag and drop', () => {
```
nawel-aa, line 190: You should test if the `change` event has been dispatched properly when we drag the tile.

src/modules/base/kanban/__tests__/kanban.test.js
```diff
@@ -0,0 +1,251 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { createElement } from 'lwc';
+import Kanban from '../kanban';
+import { ACTIONS, FIELDS, GROUP_VALUES, RECORDS } from './data';
+
+let element;
+describe('Kanban', () => {
```
nawel-aa, line 39: I think you can test the `actionclick` event too.

nawel-aa (changes requested):
See also my comment on Jira, and let me know if anything isn't clear! 🙂

src/modules/base/kanban/__tests__/kanban.test.js
AchilleStH: So, i fixed pretty much everything but the tests... I might need your help here 😵‍💫.

The problem seems to be that the tests can't compute the groups ( the `_computedGroups` array is always empty ) in the function `updateTiles()` called in the `connectedCallback()`...

This causes the `for:each` loops to iterate over an empty array in the HTML thus not rendering the groups, subgroups, tiles etc... causing most of the querySelectors to return `null` ( only happens in tests, even when all the attributes are set ).

I tried to fix it yesterday without success, have you ever encountered something similar ?

nawel-aa: I think your problem comes from the fact that you create the `computedGroups` only in the `connectedCallback()`. This works in Storybook, because the component is recreated from scratch every time we change a property value. 

In the tests (which is closer to a real case situation), the component is appended to the DOM in the `beforeEach()`, which is when the `connectedCallback()` is called. Afterwards, when you update a property, it will only go through its `set()` and the `renderedCallback()`, never through the `connectedCallback()`. The way to fix your problem, is to set a `_connected` property to `true` in the `connectedCallback()`. In each property `set()` that is used in `updateTiles()` (`groupValues`, `records`, etc.), if `_connected` is `true`, call `updateTiles()`.

Does it make sense? 🙂

AchilleStH: Works perfectly 🙌

src/modules/base/kanban/kanban.js
```diff
@@ -1653,14 +1570,14 @@ export default class Kanban extends LightningElement {
         this.dispatchEvent(
             new CustomEvent('change', {
                 detail: {
-                    id: arr[fromIndex].id,
-                    action: arr[fromIndex][this.groupFieldName],
-                    records: this.records
+                    tileId: this._draggedTile.dataset.recordIndex,
+                    droppedTo:
```
nawel-aa, line 1574: - Instead of `droppedTo`, `groupValue` would be more consistent with what we have in other components, and with `tileId`.
- Don't forget to update the JSDoc comment above, and Confluence.

src/modules/base/kanban/__tests__/kanban.test.js
```diff
@@ -248,4 +249,68 @@ describe('Kanban', () => {
...
+            });
+            tile.dispatchEvent(new MouseEvent('mousedown'));
+
+            // mousemove is handled on the kanban, not the tile
+            tile.parentElement.parentElement.dispatchEvent(
+                new MouseEvent('mousemove')
+            );
+            tile.dispatchEvent(new MouseEvent('mouseup'));
+
+            expect(handler).toHaveBeenCalled();
```
nawel-aa, line 285: You should also test that the events `detail` contain the write info, and that it bubbles/composed/cancel or not. You can see the `actionclick` test in `avatar.test.js`, for an example. This comment is also true for your `actionclick` test. 🙂

src/modules/base/kanban/kanban.html
```diff
@@ -36,439 +36,278 @@
     <lightning-spinner if:true={isLoading}></lightning-spinner>
```
nawel-aa, line 36: I missed it in my last review, but you should add an `alternative-text` to the spinners in both templates. We usually use `"Loading..."`. 🙂

src/modules/base/kanban/kanbanSubGroups.css
```diff
@@ -0,0 +1,402 @@
+.avonni-kanban__disabled_column_drag .avonni-kanban__group_header {
+    pointer-events: none;
+}
+
+.avonni-kanban__field_disabled_column_drag .avonni-kanban__summarize_wrapper {
```
nawel-aa, line 5: The declarations that are repeated between `kanban.css` and `kanbanSubGroups.css` should be added to a `shared.css` file, and then imported in both files with an `@import './shared.css'`. You can checkout `primitiveSchedulerEventOccurrence` for an example.

nawel-aa (changes requested):
Nice refactor! 👍 Your code is easier to read and the structure is clearer.
On top of the small comments here, I noticed a few bugs left that you can see in Jira. Also, in `kanban.js`, `animateSummary`, `capContainerWidth`, `capFieldHeight` and `setContainerDimensions` should not execute if `isLoading` is `true`, since they rely on elements that won't be in the DOM.

src/modules/base/kanban/kanban.js
```diff
@@ -154,9 +162,18 @@ export default class Kanban extends LightningElement {
...
      * @public
+     * @required
      */
     @api groupFieldName;
 
+    /**
+     *
+     * Name of a key of the records objects. This key needs to be present in all records objects. Its value needs to be unique to a record, as it will be used as the record identifier.
+     * @type {string}
+     * @public
```
nawel-aa, line 173: Good catch on adding `@required` to `groupFieldName`. You should also add it for `keyField`.

src/modules/base/kanban/kanban.js
```diff
@@ -905,26 +940,45 @@ export default class Kanban extends LightningElement {
      * @param {Event} event
      */
     handleActionClick(event) {
+        const recordAction = this._records.find((record) => {
+            return record[this.keyField] === event.target.dataset.keyField;
+        });
+
+        const actionName =
+            event.detail.value || event.currentTarget.dataset.name;
+        const keyField = event.currentTarget.dataset.keyField || '';
```
nawel-aa, line 949: The default value for empty `targetKeyField` or `groupValue` should be `undefined`, not an empty string.

src/modules/base/kanban/kanban.html
```diff
@@ -0,0 +1,379 @@
...
+                        onmouseup={handleGroupMouseUp}
+                    ></div>
+                    <div
+                        class="avonni-kanban__group_header"
+                        onmousedown={handleGroupMouseDown}
+                        onmouseup={handleGroupMouseUp}
+                        data-element-id="avonni-kanban__group_header"
+                        if:false={hideHeader}
+                    >
+                        <div if:true={group.headerActions}>
```
nawel-aa, line 157: Since `group.headerActions` will always be an array, it should be `if:true={group.headerActions.length}` (don't forget to update the second template too).

nawel-aa (changes requested):
Almost there. 🙂
- There is still a bug with the drag and drop (see Jira).
- `key-field="id"` needs to be added to all the examples HTML files.

nawel-aa (approved):
Everything looks good to me! Congrats! 🥳

nawel-aa (changes requested):
See my comment on Jira, there is an error with the scrolling interval.

nawel-aa (changes requested):
The tests are broken. 🙂

nawel-aa (approved):
I think we're all good! 👌

jbverger (approved):

PR #359, opened 2022-05-24
Feature abc 610 create bar code component

ivanivanov-github:
### Features
**Barcode**
- New barcode component


src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+//  * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+//  * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+//  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+//  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//  */
+
+import { LightningElement, api } from 'lwc';
+import { normalizeBoolean, normalizeString } from 'c/utilsPrivate';
+import { LIBRARY_ENCODING_VALUE, BARCODE_LIBRARY } from 'c/utilsPrivate';
```
nawel-aa, line 35: You can merge the two `c/utilsPrivate` imports.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+ * @descriptor avonni-barcode
+ * @storyId example-barcode--base
+ * @public
+ */
+export default class Barcode extends LightningElement {
+    /**
+     * The value of the barcode.
+     *
+     * @public
+     * @type {number}
```
nawel-aa, line 89: I think the value type is string, not number?

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+     * @default 300
+     */
+    @api
+    get size() {
+        return this._size;
+    }
+    set size(value) {
+        this._size = value;
```
nawel-aa, line 215: You should call `setCanvasWidth()` in `set size()` if the first render already happened, in case the size is changed after the first render (which can happen in a real life example).

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+
+    /*
+     * ------------------------------------------------------------
+     *  PUBLIC PROPERTIES
+     * -------------------------------------------------------------
+     */
+    /**
+     * Returns wether the barcode can be rendered as an svg element.
+     */
+    @api
```
nawel-aa, line 118: I don't think this getter needs to be public, you can remove the `@api` tag.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    /**
+     * The color of the background. Valid formats include color name, HEX and RGB.
+     *
+     * @public
+     * @type {string}
+     * @default #fff
+     */
+    @api
```
nawel-aa, line 132: I have identified [4 cases where you need to have a get/set](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Use-of-getter%2Fsetter-for-public-fields) for a public property.
I think `background` and other of your properties do not fit any of these case, so you can declare them at the top of the file, the same way you did for `value`. 🙂

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+    get type() {
+        return this._type;
+    }
+    set type(value) {
+        this._type = value;
+    }
+
+    /*
+     * ------------------------------------------------------------
+     * PRIVATE PROPERTIES
```
nawel-aa, line 250: The code below corresponds to private methods, not properties.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+    extractColorFromHEX(color) {
+        return color.substring(1);
+    }
+
+    /**
+     * Returns the library used to render the barcode.
+     *
+     * @returns {string} canvas
+     */
+    getBarcodeLibrary() {
```
nawel-aa, line 267: Since they don't take any argument, always return a value, and don't update the value of another property, the three `getSomething()` methods could be converted into actual private `get something()`.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+     * @returns {string} canvas
+     */
+    getLibraryEncodingValue() {
+        return LIBRARY_ENCODING_VALUE.get(this.type);
+    }
+
+    /**
+     * Calls the methods necessary to initials the barcode attributes.
+     */
+    initBarcode() {
```
nawel-aa, line 292: Since `initBarcode()` is only calling `setCanvasWidth()`, I think you can remove it and just call `setCanvasWidth()` instead.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+    /*
+     * ------------------------------------------------------------
+     * PRIVATE METHODS
+     * ------------------------------------------------------------
+     */
+    /**
+     * Sets the width for the canvas.
+     */
+    setCanvasWidth() {
+        let element = this.template.querySelector(
```
nawel-aa, line 305: This `let` could be a `const`. 🙂

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,388 @@
...
+            '[data-element-id="canvas-wrapper"]'
+        );
+        element.style.width = `${this.size}px`;
+    }
+
+    /**
+     * Switch case switches based on the library used to render the barcode. The libraries are BwipJs and JsBarCode.Each case calls the corresponding render method.
+     */
+    @api
```
nawel-aa, line 314: I don't think this method needs to be accessible by the parent component of the barcode. You can remove the `@api`.

nawel-aa (changes requested):
The component works, but of course I have comments. 😊 
You should also remove the empty `mocked-version-of-bwip.js` file.

src/modules/base/barcode/barcode.js
ivanivanov-github: You're right, it doesn't need to be accessible. However I used it to be able to call the function in the tests. Should I leave it, or not use the function in the tests?

nawel-aa: If it's only for the test, you should remove it. The `@api` is supposed to decorate only methods that it would make sense to call from outside of the component. Moreover, once we deploy it to Salesforce, we will never be able to remove or rename this method. Salesforce prevents us from deleting public properties/methods, so we have to really think twice before creating one. 😅

What you need in your tests, is to allow the component to rerender after you change its properties. If you look at other tests, this is why we always add a `return Promise.resolve()` after changing the element's properties. This will trigger the rerender, which should trigger the call to your `renderBarcode()` method. 🙂

ivanivanov-github: Got it! I've made the necessary changes!

PR #368, opened 2022-06-07
[Feat] (ABC-654) new slider component - Slider

Felino203:
### Features
Slider
- The avonni slider allows to pick a value on a range. The value attribute can be set to an array of number to allow a range style slider.
- Allows for LDS unit attributes and custom labels.
- Vertical and Horizontal styles.
- Tick marks can be displayed on the range.
- Allows for slider track and thumb customization with styling hooks (see doc).
- Value validation following LDS syntax.
- Allows for decimal steps, up to 0.001


src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+            !this.hideMinMaxValues
+        );
+    }
+
+    /**
+     * min for display in DOM
+     * @returns {number}
+     *
+     */
+    get privateMin() {
```
nawel-aa, line 777: Same comment than above, you can just rename `_computedMin` and `_computedMax` to `computedMin` and `computedMax`.

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+            }
+        }
+        return this._constraintApis;
+    }
+
+    /**
+     * Returns the current variable for styling hook of thumb radius.
+     * @returns {number}
+     */
+    get _thumbRadius() {
```
nawel-aa, line 850: The styling hook may be set on a parent component, and it will still apply to the slider. Also, the radius could be set in % or any other unit. If you want to get the width of the element, try `offsetWidth` or `getBoundingClientRect`.

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+     */
+    get _track() {
+        return this.template.querySelector('[data-element-id="track"]');
+    }
+
+    /**
+     * Returns the current variable for styling hook of track height.
+     * @returns {number}
+     */
+    get _trackHeight() {
```
nawel-aa, line 873: Same comment as above, use Javascript to get the height, not the CSS variables.

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+
+    /**
+     * Removes keyboard focus from the input element.
+     *
+     * @public
+     */
+    @api
+    blur() {
+        if (this._rendered) {
+            this.template
```
nawel-aa, line 906: You can probably use `this.template.activeElement` instead?

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+        this._computedValues.forEach((val, index) => {
+            this._computedValues[index] = Math.min(
+                Math.max(val, this._computedMin),
+                this._computedMax
+            );
+            this._computedValues[index] =
+                Math.round(this._computedValues[index] / this._step) *
+                this._step;
+        });
+        this._hideTrack = this._computedValues.length > 2 || this._hideTrack;
```
nawel-aa, line 1022: If I understand this line correctly, once the track has been hidden, it will never be unhidden, even if the value is changed.

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+    /**
+     * Calculate Pins position.
+     */
+    setPinPosition(event) {
+        let pin = this.template.querySelector(
+            `[data-group-name="pin"][data-index="0"]`
+        );
+        let pinProgress =
+            this.getPercentOfValue(
+                this._computedValues[parseInt(event.target.dataset.index, 10)]
```
nawel-aa, line 1585: - To prevent long lines, use a variable for the index.
- It's better to use `currentTarget` when possible, in case a child element triggered the event, and not the element itself. In this case it doesn't matter because nothing will ever be in the `<input>`, but keep it in mind. 🙂 
- Don't forget to use `const` instead of `let`.

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+            ) * PERCENT_SCALING_FACTOR;
+        let transformedValue =
+            this._computedValues[parseInt(event.target.dataset.index, 10)];
+        if (this._scalingFactor !== 1) {
+            transformedValue = transformedValue / this._scalingFactor;
+        }
+        pin.firstChild.firstChild.value = transformedValue;
+
+        if (!this.isVertical) {
+            pin.style.left = `calc(${pinProgress}% - ${
```
nawel-aa, line 1595: Do you have to do the calculation in the CSS? If not, it's cleaner to to it in the Javascript, since you're already setting the CSS style here.

nawel-aa (changes requested):

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+        this._computedMin = this._min * this._scalingFactor;
+        this._computedMax = this._max * this._scalingFactor;
+    }
+
+    /**
+     * If slider is closer to mouse, adds a class which puts it above the others.
+     *
+     * @param {Event} event
+     */
+    setClosestOnTop(event) {
```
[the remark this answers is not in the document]
Felino203: An inputs' thumb cannot be dragged unless it is on top of the other input (or else the same input will always be targeted y clicks). This calculation serves to set the thumb closest to the mouse on top, so that clicking on the input track will call the event on the closes thumb (<input type=range> thumb and track are linked and cannot be selected individually).

src/modules/base/slider/slider.js
Felino203: This method lets me get the computed style which still applies to slider. Also, it is the only method I know of that works to get a pseudo element attribute.

src/modules/base/slider/slider.js
```diff
@@ -0,0 +1,1709 @@
...
+        this._computedValues = [...this._value];
+        if (this._connected) {
+            this.scaleValues();
+            this.capValues();
+        }
+        this._domModified = true;
+    }
+
+    /**
+     * The variant changes the appearance of the slider. Accepted variants include standard and label-hidden.
```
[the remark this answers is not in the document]
Felino203: I dont know... the LWC slider does not have that variant and JB did not insist on him. I should ask what he thinks of this, since an inline-label variant would be nice to have.

nawel-aa (changes requested):
See my comment on Jira. 🙂

src/modules/base/slider/slider.js
```diff
@@ -1484,7 +1484,11 @@ export default class Slider extends LightningElement {
         );
         if (isNaN(parentHeight)) {
             parentHeight = this.template.host.parentElement.clientHeight;
+            if (parentHeight > 1350) {
+                return;
+            }
         }
+        console.log(parentHeight);
```
nawel-aa, line 1491: ...and don't forget to remove the console logs! 😊

nawel-aa (changes requested):

src/modules/base/slider/__examples__/vertical/vertical.html
```diff
@@ -13,6 +13,8 @@
             size="x-small"
             class="first-slider"
             style="z-index: 5; margin: 0 20px"
+            hideMinMaxValues
```
nawel-aa, line 16: The attributes must be written in kebab case, so `hide-min-max-values` instead of `hideMinMaxValues`.

nawel-aa (changes requested):

nawel-aa (changes requested):
I'm sorry I missed all the `messageWhen...`  in my previous reviews! Can you remove these ones? They don't make sense for the slider. 🙂
- `messageWhenBadInput`
- `messageWhenTypeMismatch`
- `messageWhenValueMissing`

nawel-aa (approved):
Everything's good, nice job! 👍

PR #373, opened 2022-06-15
[fix](abc-645): Remove redundant slds avatar border

lgilmet:
Fix
- Removed doubled border around avatars in a group of two.

### DIFF
--- a/src/modules/base/avatarGroup/avatarGroup.css
+++ b/src/modules/base/avatarGroup/avatarGroup.css
@@ -39,6 +39,10 @@
     cursor: pointer;
 }
 
+.slds-avatar-grouped {
+    border: none;
+}
+
 .avonni-avatar-group_x-large {
     width: 4.5rem;
     height: 4.5rem;


nawel-aa (changes requested):
It works, but it's better not to use the SLDS class directly, because once in Avonni, it could apply to other components too. 🙂
You can:
- Remove entirely `slds-avatar-grouped` and apply its `line-height` and `background` to another class.
- Add another class to the element and set the border of this one to 0.

src/modules/base/avatarGroup/avatarGroup.html
```diff
@@ -144,7 +144,61 @@
...
                 </template>
             </template>
-            <template if:true={item.showMore}>
+
+            <span
+                if:true={item.showMore}
+                key={item.key}
+                class={showMoreSectionClass}
+            >
+                <div
```
nawel-aa, line 153: We shouldn't have a `<div>` inside a `<span>`. If you want to make it inline, you can use a `<div>` with `slds-show_inline` or `slds-show_inline-block`.

src/modules/base/avatarGroup/avatarGroup.css
```diff
@@ -208,7 +208,7 @@
 }
 
 .avonni-avatar-group__avatar_color-border {
-    border-color: var(--avonni-avatar-group-avatar-stack-color-border, #ffffff);
+    border: none;
```
nawel-aa, line 211: If you remove the styling hook, we won't be able to set the border colour anymore.

nawel-aa (changes requested):
I think there's also a double border when the avatars are stacked but there are more than two avatars. 🤔

src/modules/base/avatarGroup/avatarGroup.html
lgilmet: Gotcha

src/modules/base/avatarGroup/avatarGroup.css
```diff
@@ -193,31 +218,32 @@
 .avonni-avatar-group__action-button_large {
     width: 3rem;
     height: 3rem;
+    margin-bottom: 2.6px;
```
nawel-aa, line 221: Don't forget to remove these `margin-bottom`. 🙂

nawel-aa (changes requested):
Only some forgotten styling, otherwise it works!

nawel-aa (approved):

nawel-aa (approved):
The show more button bug is solved and the popover closes as it did before. 👍

PR #376, opened 2022-06-16
[feature](abc-646): Add name to avatar group items

lgilmet:
Added 'name' property to identify an avatar in items. 
Added 'name' parameter to 'avatarclick' event. 
Added 'targetName' parameter to 'avataractionclick' event.


nawel-aa (commented):
It would be nice to add names to the example files, just so it's coherent with the doc. But aside from that, it works! 🙂

lgilmet:
Updated example names

nawel-aa (approved):

PR #377, opened 2022-06-16
[feature](abc-565): Add name to list item

lgilmet:
Add name to list item, to be returned when item is clicked.


nawel-aa (commented):
Same as #376, it would be nice to add the names to the doc examples. Aside from that, it's all good!

PR #381, opened 2022-06-21
[Feat](ABC-622) - Horizontal Activity Timeline

GenevievePMD:
### Features
Activity Timeline
- Horizontal view for activity timeline


src/modules/base/activityTimeline/activityTimeline.js
```diff
@@ -378,6 +450,28 @@ export default class ActivityTimeline extends LightningElement {
...
+    }
+
+    /**
+     * Get the div container of horizontal activity timeline
+     *
+     * @type {object}
+     */
+    get divHorizontalTimeline() {
+        return this.template.querySelector(
+            '.avonni-activity-timeline__horizontal-timeline'
```
nawel-aa, line 470: You should use a [`data-element-id` attribute](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Only-data-attributes-should-be-used-in-query-selectors) instead of a class, so your Javascript doesn't depend on your CSS.

src/modules/base/activityTimeline/activityTimeline.js
```diff
@@ -673,6 +827,48 @@ export default class ActivityTimeline extends LightningElement {
...
+     * Handle the click on an item. Dispatch the itemclick event.
+     *
+     * @param {Event} event
+     */
+    handleItemClick(event) {
+        /**
+         * The event fired when a user clicks on an item.
+         *
+         * @event
+         * @name itemClick
```
nawel-aa, line 839: - The name of the event is lowercase (`itemclick` and not `itemClick`).
- I think it should not be dispatched when the click is on the close button of the popover.

src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
```diff
@@ -65,6 +71,10 @@ describe('Activity Timeline', () => {
         expect(element.buttonShowMoreIconPosition).toBe('left');
         expect(element.buttonShowMoreLabel).toBe('Show more');
         expect(element.buttonVariant).toBe('neutral');
+        expect(element.position).toBe('vertical');
+        expect(element.horizontalTimeline).toBeUndefined();
+        expect(element.showItemPopOver).toBeFalsy();
+        expect(element.selectedItem).toBeUndefined();
```
nawel-aa, line 77: These properties are not public, so they will always be `undefined` (which is also a falsy value).

src/modules/base/activityTimeline/__tests__/activityTimeline.test.js
```diff
@@ -701,4 +711,874 @@ describe('Activity Timeline', () => {
...
+            HorizontalActivityTimeline.prototype,
+            'createHorizontalActivityTimeline'
+        );
+        element.title = 'Horizontal Activity Timeline';
+        element.items = horizontalItemsTest;
+        element.position = 'horizontal';
+
+        return Promise.resolve().then(() => {
+            const timelineContainer = element.shadowRoot.querySelectorAll(
+                '.avonni-activity-timeline__horizontal-timeline svg'
```
nawel-aa, line 728: I understand that you're targeting the `svg` element directly since it is generated by your javascript, but ideally, `.avonni-activity-timeline__horizontal-timeline` would be replaced by a [`data-element-id`](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript#Only-data-attributes-should-be-used-in-query-selectors) in case we ever change the class. 🙂
Same for all the query selectors in the tests: if you can avoid using a class, it's better.

nawel-aa (changes requested):
I didn't go too deep in the code, since it's working and I don't know d3 well enough to comment on its use. But it's very neat and organized! You did a really nice job of commenting and structuring, I'm sure it will make our life easier in the future! 👌
The only required change is the use of output-data in the popover. Otherwise, it's only small normalization rules.

nawel-aa (changes requested):

nawel-aa (approved):
I think everything is working! Congrats! 🎉

jbverger (approved):

PR #383, opened 2022-06-22
[feature] (ABC-552): Add name to pill container item

lgilmet:


nawel-aa (commented):
Same as #376, it would be nice to add the names to the doc examples. Aside from that, it's all good!

jbverger (approved):

PR #390, opened 2022-07-05
Abc 598 calendar add option to hide navigation

ivanivanov-github:
### Features
**Calendar**
- Added attribute hide-header

### DIFF
--- a/src/modules/base/calendar/__docs__/calendar.stories.js
+++ b/src/modules/base/calendar/__docs__/calendar.stories.js
@@ -63,8 +63,8 @@ export default {
                 type: { summary: 'string|string[]' }
             }
         },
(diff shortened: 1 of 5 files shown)


nawel-aa (changes requested):
You forgot a part of the task description: the `navigate` event should also be fired when the month is changed using the keyboard. It's important because the event is the most useful when the navigation is hidden, but the user can still change the month with the keyboard. 🙂

nawel-aa (changes requested):
Everything from my previous comment works! 👍 But the `navigate` event should also be dispatched when we change the year through the combobox, though.

nawel-aa (approved):
👍

PR #391, opened 2022-07-08
Abc 602 calendar add methods

ivanivanov-github:
### Features
**Calendar**
- Added public methods:  nextMonth, previousMonth, goToDate(date)
- Parameters: date: Date, String, Object - Date to navigate to. Can be a Date object or ISO8601 date string)


nawel-aa (approved):
Works well! 👌

PR #402, opened 2022-08-01
[Fix] (ABC-703) - Horizontal Activity Timeline : Fix icons on website

GenevievePMD:
### Fixes
**Horizontal Activity Timeline**
- Fix icons on website
- Change size of action icons


src/modules/base/activityTimeline/horizontalActivityTimeline.js
```diff
@@ -874,6 +1108,32 @@ export class HorizontalActivityTimeline {
...
+        if (this.areIconLibrariesReady) {
+            const libraryName = `${iconInformation.category}IconLibrary`;
+            const iconFileName = `${iconInformation.category}_${iconInformation.iconName}`;
+            template = this[`${libraryName}`][`${iconFileName}`];
+
+            if (template && template !== null) {
+                // Removing different whitespace characters and end of string
+                template = template
+                    .toString()
+                    .replaceAll(/\s{2,}/g, '')
```
nawel-aa, line 1128: `replaceAll` will cause an error when we create the next package version, because [it's only supported by Node.js >= 15](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#compatibilit%C3%A9_des_navigateurs). And you can see in `package.json` that this project needs to be compatible with Node.js >= 12.18.3.
You can use `replace(/yourRegex/g, 'the replacement')` instead.

src/modules/base/activityTimeline/horizontalActivityTimeline.js
```diff
@@ -942,6 +1214,47 @@ export class HorizontalActivityTimeline {
...
     }
 
+    /**
+     * Initialize item's popover with correct position and classes.
+     */
+    initializeItemPopover(element) {
+        let tooltipElement = d3.select(this.itemPopoverSelector);
+
+        if (
+            !tooltipElement._groups[0][0] ||
```
nawel-aa, line 1224: I missed that in my previous reviews, but I think you don't need `tooltipElement._groups[0][0] === null`, since `null` is already falsy and will pass the first part of the check. 🤔

src/modules/base/activityTimeline/horizontalActivityTimeline.js
```diff
@@ -855,6 +1031,64 @@ export class HorizontalActivityTimeline {
...
+            attributes[`${attributeName}`] = attributeValue.trim();
+        });
+
+        return attributes;
+    }
+
+    /**
+     * Extract all elements and attributes from icon's template.
+     *
+     * @return {[object, array]]}
```
nawel-aa, line 1057: 1- I think it would make more sense to return an object with two keys, than an array. It's not a list of some kind, but two different values that will always be present.
2- This `@return` type is invalid and will break the docs script on the website. Considering my first point, you would have `@return {object}` instead.

nawel-aa (changes requested):
I have small changes and suggestions, but works well whatever the path to the icons is! 👍

nawel-aa (approved):
Working well! 👌

PR #430, opened 2022-08-31
[Feat] (ABC-754):  Add more informations to popover of horizontal activity timeline

GenevievePMD:
### Features
**Horizontal activity timeline**
- Add formatted date, description and actions to item's popover

### DIFF
--- a/src/modules/base/outputData/outputData.css
+++ b/src/modules/base/outputData/outputData.css
@@ -47,16 +47,10 @@ span {
     display: initial;
 }
 
-
 .avonni-output-data__label {
     font-weight: var(--avonni-output-data-label-font-weight, 'normal');
-    width: var(--avonni-output-data-label-width, '100%');
-    padding-right: var(--avonni-output-label-data-padding-right, 0);
-     /* slds-text-color_weak */
-    color: #444444;
-    /* slds-truncate */
-    max-width: 100%;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
+}
+
+.avonni-output-data__label_inline {
+    width: 30%;
 }
--- a/src/modules/base/outputData/outputData.css
+++ b/src/modules/base/outputData/outputData.css
@@ -46,3 +46,17 @@ span {
 .slds-form-element__control .slds-checkbox {
     display: initial;
 }
+
+
+.avonni-output-data__label {
+    font-weight: var(--avonni-output-data-label-font-weight, 'normal');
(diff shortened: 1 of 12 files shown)


src/modules/base/outputData/outputData.css
```diff
@@ -46,3 +46,17 @@ span {
 .slds-form-element__control .slds-checkbox {
     display: initial;
 }
+
+
+.avonni-output-data__label {
+    font-weight: var(--avonni-output-data-label-font-weight, 'normal');
+    width: var(--avonni-output-data-label-width, '100%');
+    padding-right: var(--avonni-output-label-data-padding-right, 0);
+     /* slds-text-color_weak */
```
nawel-aa, line 55: You shouldn't replace SLDS classes by custom CSS if you can avoid it. 🙂

nawel-aa (changes requested):
You should update the documentation example too, so it is similar to the storybook example. 🙂

src/modules/base/outputData/outputData.js
```diff
@@ -332,4 +335,16 @@ export default class OutputData extends LightningElement {
 
         this.normalizedTypeAttributes = normalizedTypeAttributes;
     }
+
+    /**
+     * Set styling hooks of output data's label if variant is label-inline.
+     */
+    setLabelStylingHook(){
```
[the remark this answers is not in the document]
nawel-aa: **Some clarification after talking with JB:**
Instead of having a `setLabelStylingHook()`, you should keep the `computedLabelClass()` that we had before. When the variant is `label-inline`, you should add a custom class and a SLDS padding class:
```
get computedLabelClass() {
    return classSet('slds-item_label slds-text-color_weak slds-truncate')
        .add({
            'slds-assistive-text': this.variant === 'label-hidden',
            'slds-p-right_small avonni-output-data__label_inline': this.variant === 'label-inline'
        })
        .toString();
}
```

In your CSS, the custom class will contain the width:
```
.avonni-output-data__label_inline {
    width: 30%;
}
```

src/modules/base/outputData/outputData.css
nawel-aa: **Some clarification, after talking with JB:**
- You should keep `slds-text-color_weak` and `slds-truncate`.
- This CSS class should only contain `font-weight: var(--avonni-output-data-label-font-weight, 'normal')` (see the comment in Javascript about the width and padding).

nawel-aa (approved):

PR #431, opened 2022-09-01
Feature abc 610 create barcode component

lgilmet:
Refactor barcode component


src/modules/base/barcode/__tests__/barcode.test.js
```diff
@@ -0,0 +1,156 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { createElement } from 'lwc';
+import Barcode from 'c/barcode';
+
+let element;
+describe('Barcode', () => {
```
nawel-aa, line 37: These tests will always pass, because they're setting a value, and then checking that the value has been set. 😊
I get that the component rely heavily on a library, so you can't test the end result. But you should check if the library is called with the right parameters each time. Something like so:

```
const handler = jest.fn();
bwipjs.toCanvas = handler;
expect(handler).toHaveBeenCalledWith('whateverarguments');
```

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+//  * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+//  * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+//  * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+//  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//  */
+
+import { LightningElement, api } from 'lwc';
+import { normalizeBoolean } from 'c/utilsPrivate';
+import bwipjs from 'bwip-js';
+import { normalizeString } from '../utilsPrivate/normalize';
```
nawel-aa, line 36: `utilsPrivate` needs to be imported from the `c` namespace.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+        'center-right',
+        'center-justify',
+        'bottom-left',
+        'bottom-center',
+        'bottom-right',
+        'bottom-justify'
+    ],
+    default: 'bottom-center'
+};
+const TEXT_X_ALIGN = ['left', 'center', 'right', 'justify'];
```
nawel-aa, line 58: To follow our convention, instead of having an array of valid values, and a default value written directly in the property, your constants should use an object with two keys: valid and default.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+     * @public
+     * @type {boolean}
+     * @default true
+     */
+    @api
+    get checksum() {
+        return this._checksum;
+    }
+    set checksum(value) {
+        this._checksum = normalizeBoolean(value);
```
nawel-aa, line 243: All the properties that are used in `barcodeParameters()` should call `renderBarcode()` if the component is already connected. Their change only won't trigger a rerender, because they are not used in the DOM.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+     * @public
+     * @type {number}
+     */
+    @api
+    get height() {
+        return this._height;
+    }
+    set height(value) {
+        const numValue = parseInt(value, 10);
+        if (numValue != null) {
```
nawel-aa, line 258: - Try not using loose equality (use `!==` instead of `!=`).
- I think you should check if it is `NaN`, not if it is `null`. 🤔  And if it is invalid (null, undefined, a string, whatever), you should reset the value to an empty value, instead of keeping the old value.

(same comment for the width)

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+    @api
+    get hideValue() {
+        return this._hideValue;
+    }
+    set hideValue(value) {
+        this._hideValue = normalizeBoolean(value);
+    }
+
+    /**
+     * The position of the displayed value. Accepted values are top-left, top-center, top-right, top-justify, center-left, center-center, center-right, center-justify, bottom-left, bottom-center, bottom-right, bottom-justify. Defaults to bottom-center.
```
nawel-aa, line 279: You don't need to say "Defaults to..." in the properties description, because that's what the `@default` is for. It's only for secondary objects that you have to add it at the end, because we don't have an `@default`.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+     *
+     * @public
+     * @type {text}
```
nawel-aa, line 282: `text` is not a valid type, it should be `string`

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+    @api
+    get textAlignment() {
+        return this._textAlignment;
+    }
+    set textAlignment(value) {
+        this._textAlignment = normalizeString(value, {
+            fallbackValue: TEXT_ALIGNMENT.default,
+            valid: TEXT_ALIGNMENT.valid
+        });
+        if (this._textAlignment) {
```
nawel-aa, line 294: There will always be a `textAlignment`, since there is a normalize.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+            const replace1 = this._textAlignment.replace('bottom', 'below');
+            const replace2 = replace1.replace('top', 'above');
+            const outputAlignment = replace2.split('-');
+            this.textXAlign = normalizeString(outputAlignment[1], TEXT_X_ALIGN);
```
nawel-aa, line 298: This is not the accepted format for `normalizeString`. The second argument should be an object. You can check how `normalizeString` works in the utils.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,426 @@
...
+        }
+
+        return params;
+    }
+
+    /**
+     * Sets the width for the canvas.
+     */
+    get barcodeStyle() {
+        return `${
```
nawel-aa, line 381: - Try not to make too big computations in an interpolation. It would be easier to read if you used two variables.
- What if the width and height are `undefined`, instead of `null`? Why not just check if they are present? I don't think we want a width/height of 0 anyway.
- This get() and the next one got lost in the methods section.

nawel-aa (changes requested):

src/modules/base/barcode/barcode.js
lgilmet: As we discussed, != null is also equal to != undefined :)

src/modules/base/barcode/barcode.js
lgilmet: Yes, that was not intended.

src/modules/base/barcode/barcode.js
lgilmet: 🤦

src/modules/base/barcode/barcode.js
lgilmet: It was bugging when I called renderBarcode in the prop setters, so I call it only after the initial render.

nawel-aa (changes requested):
Nice tests! 🙂
There's just a problem with the error that never goes away once it's here. And a suggestion about default values.

src/modules/base/barcode/barcode.html
```diff
@@ -0,0 +1,43 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+-->
+
+<template>
+    <div if:true={validCode} class="slds-text-align_center">
```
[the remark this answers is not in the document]
lgilmet: I need to start testing props from the inside of the components instead of using Storybook.

src/modules/base/barcode/barcode.js
```diff
@@ -0,0 +1,495 @@
...
+        this._value = normalizeString(codeValue, { toLowerCase: false });
+        this.rerenderBarcode();
+    }
+
+    /**
+     * The width of the barcode in pixels. The width defaults to 100%.
+     *
+     * @public
+     * @type {number}
+     * @default 100%
```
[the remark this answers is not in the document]
lgilmet: I need to change an empty string to undefined because !isNaN('') = true, because '' is a number lol
 ```
       value = value !== '' ? value : undefined;
        this._height = !isNaN(value) ? `${value}px` : value;
```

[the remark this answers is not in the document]
Comment (nawel-aa) [reply src/modules/base/barcode/barcode.js:374]: - Good catch, my example was wrong, I missed the empty values cases! 🤦‍♀️  It's the same problem with `null`, and since we don't want a width of 0, you can probably use `value && !isNaN(value)`?
- My example was wrong twice, because if the value is invalid, we want to set the width back to 100% (you're doing it for `''` here, but not for `undefined` and `null`). 🙂
- You shouldn't change the value of a variable you didn't create (here `value`), it could cause a problem upstream.

File: src/modules/base/barcode/barcode.js [hunk 13]
```diff
@@ -33,6 +33,7 @@
 import { LightningElement, api } from 'lwc';
 import { normalizeBoolean, normalizeString } from 'c/utilsPrivate';
 import bwipjs from 'bwip-js';
+import { classSet } from '../utils/classSet';
```
nawel-aa, line 36: A little autocompleted bad import here.

nawel-aa (changes requested):

src/modules/base/barcode/barcode.js
lgilmet: Thanks

src/modules/base/barcode/barcode.js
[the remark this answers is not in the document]
lgilmet: So I made it so that undefined, null and '' are normalized to '' for height (because passing undefined to style.maxHeight does not change the height) 
and '100%' for width.

nawel-aa (approved):
Everything works well! 👍

jbverger (approved):

PR #435, opened 2022-09-08
Feature abc 699 add items per panel to carousel

lgilmet:
Feature: 
- Add responsive number of items per panel.


src/modules/base/carousel/__docs__/data.js
```diff
@@ -36,7 +36,7 @@ const items = [
         name: '1',
         title: 'Visit App Exchange',
         description: 'Extend Salesforce with the #1 business marketplace.',
-        imageAssistiveText: 'Appy',
```
nawel-aa, line 39: Appy is the name of [Salesforce mascot](https://appexchange.salesforce.com/appxContentListingDetail?listingId=a0N3A00000EcsHZUAZ), so this was the right assistive text. 🙂 No need to change it in all examples either.

src/modules/base/carousel/__tests__/carousel.test.js
```diff
@@ -413,6 +414,33 @@ describe('Carousel', () => {
         });
     });
 
+    it('Carousel: small items per panel', () => {
+        element.smallItemsPerPanel = 3;
+        element.items = items;
+
+        return Promise.resolve().then(() => {
+            expect(element.smallItemsPerPanel).toBe(3)
```
nawel-aa, line 422: These tests will always pass! You set a property and then check if it has been set.

Since the number of items depends on the carousel container `offsetWidth`, you can try to mock it. Here is an example of a test in the builder repo, where I mocked the `scrollHeight` of a `<ul>`, for help:
```
const list = element.shadowRoot.querySelector(
    '[data-element-id="ul-listbox"]'
);
jest.spyOn(list, 'scrollHeight', 'get').mockImplementation(
    () => 500
);
```

src/modules/base/carousel/carousel.js
```diff
@@ -297,9 +315,7 @@ export default class Carousel extends LightningElement {
                 actions: item.actions || []
             });
         });
-        if (this._initialRender) {
-            this.initCarousel();
-        }
+        this.initCarousel();
```
nawel-aa, line 318: You need the component to have rendered for `initCarousel()` to be called.

src/modules/base/carousel/carousel.js
```diff
@@ -324,9 +340,87 @@ export default class Carousel extends LightningElement {
...
+     * @type {number}
+     * @public
+     */
+    @api
+    get smallItemsPerPanel() {
+        return this._smallItemsPerPanel;
+    }
+
+    set smallItemsPerPanel(value) {
+        this._smallItemsPerPanel = Number(
```
nawel-aa, line 360: I know that's what was used for `itemsPerPanel`, but it's a weird way to check if a number is valid. Moreover, if the number is invalid, the property will be set to 0, which is itself invalid.

You can just `parseInt()` it, and check if it is greater than 10 or lesser than 1. If so, set it to `undefined`. And since it's a check you have to make in all the `set somethingItemsPerPanel()`, you can make a method out of it.

src/modules/base/carousel/carousel.js
```diff
@@ -324,9 +340,87 @@ export default class Carousel extends LightningElement {
...
+            normalizeString(
+                typeof value === 'number' ? value.toString() : value,
+                {
+                    validValues: ITEMS_PER_PANEL
+                }
+            )
+        );
+
+        this.carouselIsResponsive = true;
```
nawel-aa, line 369: What if `undefined` or an invalid value was just given? `carouselIsResponsive` should not be set to `true` by default. A getter would probably be more appropriate, since the value depends on several properties.

src/modules/base/carousel/carousel.js
```diff
@@ -324,9 +340,87 @@ export default class Carousel extends LightningElement {
...
+        this.columnsCount.medium = this._mediumItemsPerPanel;
+        this.computeItemsPerPanel();
+    }
+
+    /**
+     * Number of items to be displayed at a time in the carousel. Maximum of 10 items per panel.
```
nawel-aa, line 401: ...when the component is 1024px wide or more?

src/modules/base/carousel/carousel.js
```diff
@@ -165,6 +178,11 @@ export default class Carousel extends LightningElement {
                 this.play();
             }
         }
+        if (!this.resizeObserver) {
```
nawel-aa, line 181: If there was a resize observer and it is now useless, you should disconnect it. You should also disconnect it in the `disconnectedCallback()`, to make sure there is no memory leak. See pill container for an example.

src/modules/base/carousel/carousel.js
```diff
@@ -662,13 +777,53 @@ export default class Carousel extends LightningElement {
...
 
+    computeItemsPerPanel() {
+        if (!this.carouselContainer) {
+            return;
+        }
+
+        if (!this.carouselIsResponsive) {
+            this.currentItemsPerPanel = this.columnsCount.default;
+            return;
+        }
```
nawel-aa, line 788: I don't think you need this check, since the if/elseif/else below should work anyway if there is no specification for the other sizes.

src/modules/base/carousel/__examples__/responsiveItemsPerPanel/withoutPanelNavigationWithFiveItemsPerPanel.html
```diff
@@ -0,0 +1,12 @@
+<template>
```
nawel-aa, line 1: The name of this file and the JS file should be updated.

nawel-aa (changes requested):

src/modules/base/carousel/carousel.js
lgilmet: The property carouselIsResponsive was not useful enough to remain.

src/modules/base/carousel/carousel.js
lgilmet: I now check in normalizeItemsPerPanel. I use it for all but default items per panel because that's the only one with a default.

src/modules/base/carousel/carousel.js
lgilmet: I realized that too.

src/modules/base/carousel/carousel.js
lgilmet: Better this way? 
if (this._connected) {

src/modules/base/carousel/__examples__/responsiveItemsPerPanel/withoutPanelNavigationWithFiveItemsPerPanel.html
lgilmet: I axed this example.

src/modules/base/carousel/__tests__/carousel.test.js
lgilmet: It does not trigger the resizeObserver so I added small, medium and large items per panel to the untested properties.

nawel-aa (approved):
All seems good! 👌

jbverger (approved):

PR #466, opened 2022-09-28
[fix] (abc-539): Check if list can be sorted

lgilmet:
Make sure keyboard events are sent event if there is more than one column and list is not sortable.
Make overflow visible on single-line.
Delete repeating lines.

### DIFF
--- a/src/modules/base/list/list.js
+++ b/src/modules/base/list/list.js
@@ -800,7 +800,8 @@ export default class List extends LightningElement {
         )
             .add({
                 'slds-grid_vertical': this._currentColumnCount === 1,
-                'slds-wrap': this._currentColumnCount > 1,
+                'slds-wrap':
+                    this._currentColumnCount > 1 && this.variant === 'base',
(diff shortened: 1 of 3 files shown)


src/modules/base/list/list.js
```diff
@@ -1614,13 +1616,9 @@ export default class List extends LightningElement {
...
         }
 
-        if (this._keyboardDragged) {
-            this._keyboardDragged = false;
-            return;
-        }
-
         // Stop dragging if the click was on a button menu
         if (
+            (this._currentColumnCount > 1 && this.variant !== 'base') ||
```
nawel-aa, line 1621: You need an `||`, not an `&&`. Otherwise single-line with one item will be sortable.

nawel-aa (changes requested):

src/modules/base/list/list.js
lgilmet: Nice catch

nawel-aa (approved):
🎉

nawel-aa (approved):

PR #489, opened 2022-10-06
[fix](ABC-763): Activity Timeline - Fix the test error on package publication

GenevievePMD:
### Fixes
**Activity Timeline**
- Fix the test error on package publication (line was commented)


nawel-aa (approved):
The test seems to pass whatever the time zone. 🙂

PR #502, opened 2022-11-03
[Feature] (ABC-836): Input pen added background color attribute

Felino203:
### Features
InputPen
- Input pen added background color attribute

### DIFF
--- a/src/modules/base/inputPen/inputPen.js
+++ b/src/modules/base/inputPen/inputPen.js
@@ -242,7 +242,7 @@ export default class InputPen extends LightningElement {
      */
 
     /**
-     * Background color of the canvas.
+     * Defines the color of the background
      *
      * @type {string}
      * @public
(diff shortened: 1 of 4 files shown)


src/modules/base/inputPen/inputPen.js
```diff
@@ -989,14 +1018,22 @@ export default class InputPen extends LightningElement {
...
             this.canvasInfo.canvasElement.height
         );
-        this._backgroundCtx.globalAlpha =
-            parseInt(this._backgroundColor.slice(7), 16) / 255;
+        if (this.backgroundColor.length > 7) {
+            this._backgroundCtx.globalAlpha =
+                parseInt(this._backgroundColor.slice(7), 16) / 255;
+        } else {
+            this._backgroundCtx.globalAlpha = 1;
+        }
```
nawel-aa, line 1036: I can't test it, but I think this way of getting the alpha value won't work for the `rgba()` format. A `RegEx` is probably the way to go.

nawel-aa (changes requested):
Right now the component crashes when I try to set a background color. I'm assuming it's because the `handleChangeEvent()` called in the `set backgroundColor()` needs the component to be rendered, but I'll let you check.

nawel-aa (approved):
Works well! 👍

PR #598, opened 2023-05-23
feature-ABC-966-image-compare-slider

Jusolijus17:
### Features
Image
- Slide to compare two images


src/modules/base/image/image.js
```diff
@@ -431,6 +749,347 @@ export default class Image extends LightningElement {
...
+            compareElements.compareImg.style.height = `${posY}px`;
+        }
+    }
+
+    /**
+     * Disable the slider dragging after releasing mouse button.
+     */
+    handleCompareMouseUp() {
+        this._isDragging = false;
+    }
```
nawel-aa, line 1025: It would be nice if we could click anywhere in the whole window to release.
Since you only have a property to change, the neater way to do it would be to add the listener in `handleCompareMouseDown()` and remove it inside `handleCompareMouseUp()`. You can see an example of what I mean in the flow components repo, in `propertyEditorDialogLeftPanel` > `handleObjectFieldItemMouseDown()`.

nawel-aa (changes requested):
**Fixes:**
- We shouldn't see the magnifier in this pull request, since this feature was made in another branch. Please remove anything regarding the magnifier, and merge the master branch in this branch to resolve the remaining conflicts. 🙂
- Right now the tests don't pass.

**Documentation:**
- The compare attributes need to be documented. You need to add a `@typedef` block in the `image.jsdoc.js` file. See [Documenting Objects](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc) in our documentation.

**Improvements:** 
- Can you make it work on touch screens?
- Can you make it work with the keyboard?

src/modules/base/image/image.js
```diff
@@ -431,6 +749,347 @@ export default class Image extends LightningElement {
         if (img) {
             this._imgElementWidth = img.clientWidth;
             this._imgElementHeight = img.clientHeight;
+            if (this.magnifier) {
+                this.handleMagnifier(img);
+            }
+            if (this.compareSrc !== '') {
```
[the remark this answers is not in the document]
Jusolijus17: No, we don't, because in that case, we don't want to show the compare slider with an empty compare image. That's why I ensure that `compareSrc` isn't an empty string. To simplify the code, I changed the condition from `if (this.compareSrc !== '')` to `if (this.compareSrc)`.

src/modules/base/image/image.html
```diff
@@ -66,5 +66,62 @@
...
+            </div>
+            <div class={computedSliderClass} data-element-id="compare-slider">
+                <div
+                    class="
+                        avonni-image__compare-slider-handle
+                        slds-is-absolute
+                        slds-align_absolute-center
+                    "
+                    data-element-id="compare-slider-handle"
+                    aria-role="button"
```
nawel-aa, line 101: It's `role="button"`, not `aria-role="button"`. It's my bad, I see that I wrote it wrong in the subtask. 🤦‍♀️ 
And since we're improving accessibility, may as well add: 
- [`aria-pressed`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed) to the `<div>` to indicate whether the handle is being dragged or not.
- `alternative-text` attributes to the `lightning-icons`, indicating what to do to move the cursor ("Press left to show the original image" / "Press right to show the compared image", for example?).

This is a very visual feature, so it may seem useless to make it accessible to screen readers. But as soon as we make it focusable, we can't hide it to screen readers. So we may as well explain its purpose, instead of having a mysterious toggle button. 🙂

src/modules/base/image/image.js
```diff
@@ -612,13 +678,234 @@ export default class Image extends LightningElement {
...
+            event.preventDefault();
+            const handleMouseDown = () => {
+                this._isDraggingCompareCursor = false;
+                window.removeEventListener('mousedown', handleMouseDown);
+            };
+            window.addEventListener('mousedown', handleMouseDown);
+            if (event.target === handle) {
+                this._isDraggingCompareCursor = !this._isDraggingCompareCursor;
+            }
+        } else if (this._isDraggingCompareCursor && event.target === handle) {
```
nawel-aa, line 772: Why check if the event target is the handle? Since the event listener is placed directly on the handle, and nothing inside of it is focusable, it should always be the case.

src/modules/base/image/image.js
```diff
@@ -612,13 +678,234 @@ export default class Image extends LightningElement {
...
+            slider.style.transition = 'all 0.15s ease-out';
+            compareImg.style.transition = 'all 0.15s ease-out';
+            if (event.key === 'ArrowLeft' || event.key === 'ArrowUp') {
+                if (this.compareAttributes.orientation === 'horizontal') {
+                    if (numericValue > 20) {
+                        slider.style.left = `${numericValue - 20}px`;
+                        compareImg.style.width = `${numericValue - 20}px`;
+                    } else {
+                        slider.style.left = '0px';
+                        compareImg.style.width = '0px';
```
nawel-aa, line 798: By convention, if we can make it simpler, we do: "0px" is equal to 0. 😊

src/modules/base/image/image.js
```diff
@@ -612,13 +678,234 @@ export default class Image extends LightningElement {
...
+        const overlay = this.template.querySelector(
+            '[data-element-id="compare-overlay-container"]'
+        );
+        if (
+            this.compareAttributes.showLabelsOnHover &&
+            this.compareAttributes.originalLabel &&
+            this.compareAttributes.compareLabel
+        ) {
+            overlay.classList.remove('slds-hide');
+        }
```
nawel-aa, line 906: I think you can do the hover using CSS only. If `this.compareAttributes.showLabelsOnHover` is `true`, you can add a class in the `computedImageClass`, let's say `avonni-image__container_compare-labels-on-hover`. In your CSS, you can then have something like:

```
.avonni-image__container_compare-labels-on-hover .avonni-image__compare-overlay {
     display: none;
}
.avonni-image__container_compare-labels-on-hover:hover .avonni-image__compare-overlay {
     display: block;
}
```

I haven't tested it, but I'm guessing it should work. 😊

src/modules/base/image/image.js
```diff
@@ -629,13 +916,21 @@ export default class Image extends LightningElement {
         if (img) {
             this._imgElementWidth = img.clientWidth;
             this._imgElementHeight = img.clientHeight;
+            if (this.magnifier) {
+                this.handleMagnifier(img);
+            }
+            if (this.compareSrc) {
+                setTimeout(() => {
+                    this.handleCompareSlider(img);
+                }, 50);
```
nawel-aa, line 925: What you actually want to do is wait for the image to be rendered. It may take less or more than 50ms, so a timeout is not ideal. What you can use is [requestAnimationFrame()](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) instead.

src/modules/base/image/image.js
```diff
@@ -1268,4 +1014,112 @@ export default class Image extends LightningElement {
...
+        ) {
+            slider.style.top = `${posY}px`;
+            compareImg.style.height = `${posY}px`;
+        }
+    }
+
+    /**
+     * Initiate the compare slider.
+     */
+    handleCompareSlider(img) {
```
nawel-aa, line 1059: I think this method should be named something like `initCompareSlider()`, since it is not handling an event.

nawel-aa (changes requested):
Almost there, nice keyboard support! 👌
Aside from the few comments I had, the file should be reordered in alphabetical order, per section (public properties together, private properties, private methods, event handlers).

src/modules/base/image/__examples__/compare/compare.js
```diff
@@ -0,0 +1,11 @@
+import { LightningElement } from 'lwc';
+
+export default class Compare extends LightningElement {
+    compareAttributes = {
+        moveOn: 'click',
+        orientation: 'horizontal',
+        originalLabel: 'Before',
+        compareLabel: 'After',
+        showLabelsOnHover: false
+    };
```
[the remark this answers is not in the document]
Jusolijus17: Okay, but now that I have set `originalLabel` and `compareLabel` to empty strings by default, wouldn't it be better if I assign a value to these two for demonstration purposes?

src/modules/base/image/image.js
```diff
@@ -93,12 +93,28 @@ const MOVE_ON_OPTIONS = {
...
+    @api compareSrc;
+
     _aspectRatio;
     _compareAttributes = {
         orientation: COMPARE_ORIENTATION.default,
         moveOn: MOVE_ON_OPTIONS.default,
-        originalLabel: 'Before',
-        compareLabel: 'After',
+        originalLabel: '',
+        compareLabel: '',
```
nawel-aa, line 117: You can just remove the two keys from the default value. No need to set them to empty strings.

nawel-aa (changes requested):
Your fixes are good! But I noticed a few small bugs described in more details in the task. 🙂

nawel-aa (changes requested):

nawel-aa (approved):
Nice job, looks good to me! 🎉

PR #599, opened 2023-05-23
feature-ABC-965-image-magnifier-option

Jusolijus17:
### Features
Image
- Magnifier


src/modules/base/image/__tests__/image.test.js
```diff
@@ -166,10 +174,14 @@ describe('Image', () => {
...
         return Promise.resolve().then(() => {
-            const img = element.shadowRoot.querySelector(
-                '[data-element-id="img"]'
+            const container = element.shadowRoot.querySelector(
+                '[data-element-id="container"]'
+            );
+            const classes = container.className.split(' ');
+            expect(classes).toContain(
+                'avonni-image',
+                'avonni-image_float-left'
```
nawel-aa, line 183: Instead of using `split()` on `className` you can use `container.classList.toContain()`. If you are using this strategy (only testing if the specific class is added), you don't need to check if `avonni-image` has been added, since it is always present. The advantage of using `className` over `classList`, is that we are sure that exactly these classes are present, and not one more.

src/modules/base/image/image.js
```diff
@@ -354,13 +374,103 @@ export default class Image extends LightningElement {
...
+    }
+
     /**
      * Computed Image class styling.
      *
      * @type {string}
      */
     get computedImageClass() {
-        return classSet('avonni-image')
+        return classSet('avonni-image avonni-image_container')
```
nawel-aa, line 473: You don't need to add `avonni-image_container`, you can use the already existing `avonni-image` (or rename it), and add the new CSS you need to it.

src/modules/base/image/__tests__/image.test.js
```diff
@@ -483,4 +525,294 @@ describe('Image', () => {
...
+            expect(magnifier.style.width).toBe('150px');
+            expect(magnifier.style.height).toBe('150px');
+            expect(element.magnifierAttributes.horizontalOffset).toBe(0);
+            expect(element.magnifierAttributes.verticalOffset).toBe(0);
+        });
+    });
+
+    describe('Image: Magnifier types', () => {
+        it('Image: Standard magnifier mousemove', () => {
+            element.src = src;
```
nawel-aa, line 584: If you really want to test the magnifier mouse move, you can add a `clientX` and `clientY` position to the `mousemove` event, and mock the implementation of `getBoundingClientRect()` on the image. Otherwise, you are not really testing the magnifier, just that something happens when we hover the image.

For an example of what I mean, you can checkout "Pill container: reorder event, from single-line collapsed popover to visible pill" in `pillContainer.test.js`. You'll see the mocked implementation of `getBoundingClientRect()` on `items[0]` and the custom event `clientX` setting.

nawel-aa (changes requested):
Very neat subtest sections! I didn't know we could do that, and we will definitely start doing it everywhere. And overall, nice first big feature, congrats! 😃 My comments are mostly about conventions and refactoring, so good job! I only noticed one bug, that I mentioned on Jira.

Three more things on documentation and formatting:
- The properties and methods you added should be sorted in alphabetical order, and placed in each of their respective section. You can see the "File order" section in [our documentation](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1587019795/Javascript).
- It's a big enough feature that it deserves a story (like `Base`, `BaseSmall`, etc.) in the stories file. If you create a new story, you also have to create a new example component in `__examples__`, using exactly the same configuration as the story.
- I updated the [Confluence page](https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/363593765/Image) with the new attributes and the styling hooks, but in the future, you can do it. 🙂

Two suggestion of improvements:
- Could it be easily adapted on touch screens, using [touch events](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent#touch_event_types)? Otherwise, it will be another task.
- The functions used only for the magnifier can be extracted in another javascript file (`autoPositionMagnifier()`, `standardMagnifier()`, etc.).

src/modules/base/image/image.js
```diff
@@ -55,6 +56,15 @@ const LAZY_LOADING_VARIANTS = {
 const CROP_POSITION_X_DEFAULT = '50';
 const CROP_POSITION_Y_DEFAULT = '50';
 
+const MAGNIFIER_TYPES = {
+    valid: ['none', 'standard', 'inner', 'follow'],
+    default: 'none'
```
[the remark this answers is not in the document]
Jusolijus17: I decided to add the `none` option because you cannot deselect an option after choosing one from the drop-down menu. So, if you mistakenly clicked on `standard`, you wouldn't be able to disable the magnifier afterwards by reselecting `Choose an option...`. You would need a `none` option or an on/off switch for the magnifier.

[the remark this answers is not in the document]
nawel-aa: You can add a "none" option to the Storybook `magnifierType` argument options, if you want.
Storybook is only used during the development process, to make it easier to test what we're doing. In practice, the magnifier type will be passed by a parent component (through an HTML `magnifier-type` attribute, same as you do it in the storybook wrapper). And when we will display this option in our own property editor, in Salesforce, we will use our combobox, where we can clear the selected value: 

Screenshot 2023-05-29 at 14 52 54

src/modules/base/image/image.css
```diff
@@ -67,3 +67,22 @@
 .avonni-image_display-block {
     display: block;
 }
+
+.avonni-image_magnifier {
```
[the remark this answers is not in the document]
Jusolijus17: I changed most of the custom classes to slds classes, but I feel like adding and removing a class (`slds-hide` and `slds-show`) each time only to show and hide an element is more complicated than simply using `element.style.display = 'block/none'`. Let me know what you think.

src/modules/base/image/image.html
```diff
@@ -43,21 +43,23 @@
...
-            class="avonni-image_magnifier"
+            class="avonni-image_magnifier slds-scrollable_none slds-is-absolute"
             style={computedMagnifierStyle}
         >
             <img
                 src={src}
-                alt="test"
+                class="avonni-image_fit-content"
                 sizes={sizes}
                 style={computedMagnifiedImgStyle}
```
nawel-aa: I hadn't seen this `alt`, but it's a good idea to [make the magnifier invisible to screen readers](https://www.w3.org/WAI/tutorials/images/decorative/), by giving it an empty `alt=""`.
Unless you think it's useful to show it to screen readers, in which case I would add an `alternativeText` key to the `magnifier-attributes`.

nawel-aa (changes requested):
Really nice job, I have nothing to add! 👌
My comments are documentation related. Aside from these 3-minute fixes, it's good to go!

nawel-aa (approved):
🚀

PR #609, opened 2023-06-15
Bugfix abc 984 image magnifier option

Jusolijus17:
### Fixes
**Image**
- Fixing small issues


nawel-aa (changes requested):
One of the tests is failing. Can you check if it should still pass, or update it otherwise?

src/modules/base/image/image.js
```diff
@@ -347,15 +347,19 @@ export default class Image extends LightningElement {
...
-        )
-            ? normalizedAttributes.zoomFactor
-            : DEFAULT_ZOOM_FACTOR;
+        if (
+            normalizedAttributes.zoomFactor &&
+            !isNaN(normalizedAttributes.zoomFactor)
+        ) {
+            this._magnifierAttributes.zoomFactor =
+                normalizedAttributes.zoomFactor;
+        }
```
nawel-aa, line 362: Here it is impossible to reset the `zoomFactor` and the `smoothMove` keys. If magnifier attributes were passed, and then removed, their value would remain.

src/modules/base/image/image.css
```diff
@@ -55,6 +55,7 @@
     border-width: var(--avonni-image-magnifier-sizing-border, 1px);
     border-style: var(--avonni-image-magnifier-styling-border, solid);
     border-color: var(--avonni-image-magnifier-color-border, #ccc);
+    border-radius: var(--avonni-image-magnifier-radius-border, 0);
```
nawel-aa, line 58: Good idea! But if you add a styling hook, you need to document it in `image.jsdoc.js` and [Confluence](https://avonnicreator.atlassian.net/wiki/spaces/ABC/pages/363593765/Image). 😊

nawel-aa (changes requested):

src/modules/base/image/image.css
Jusolijus17: Yes I remembered right after requesting your review🤦‍♂️

nawel-aa (approved):

PR #612, opened 2023-06-19
[feat] (ABC-986): added disabled option

Jusolijus17:
### Features
**Visual Picker Link**
- Added disabled option

### DIFF
--- a/src/modules/base/visualPickerLink/visualPickerLink.js
+++ b/src/modules/base/visualPickerLink/visualPickerLink.js
@@ -36,6 +36,7 @@ import { classSet } from 'c/utils';
 import visualPickerLink from './visualPickerLink.html';
 import visualPickerLinkInfoOnly from './visualPickerLinkInfoOnly.html';
 
+const DEFAULT_DISABLED = false;
 const ICON_POSITIONS = { valid: ['left', 'right'], default: 'left' };
 
 /**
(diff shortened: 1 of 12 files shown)


src/modules/base/visualPickerLink/visualPickerLink.js
```diff
@@ -36,6 +36,7 @@ import { classSet } from 'c/utils';
 import visualPickerLink from './visualPickerLink.html';
 import visualPickerLinkInfoOnly from './visualPickerLinkInfoOnly.html';
 
+const DEFAULT_DISABLED = false;
```
nawel-aa, line 39: I see you got used to declaring the default values in global constants (which is good!). But for `false`, we don't store it in a variable. You can directly write `_disabled = false`. 🙂
Otherwise, in some files we would end up with too many constants with the same `false` value.

nawel-aa (commented):
Everything works well! I only have one comment on convention.

src/modules/base/visualPickerLink/visualPickerLink.js
Jusolijus17: Yes, that makes sense, but honestly, I copied that from `VisualPicker`, which declares this constant for some reason.

nawel-aa: Aaah, that makes sense, it's one of our oldest components, that doesn't always follow good practices. :)

nawel-aa (approved):
👌

PR #614, opened 2023-06-22
[refactor] (ABC-989): removed <a> tag when disabled

Jusolijus17:
### Fixes
**Visual Picker Link**
- Refactor disabled option

### DIFF
--- a/src/modules/base/visualPickerLink/visualPickerLinkInfoOnly.css
+++ b/src/modules/base/visualPickerLink/visualPickerLinkInfoOnly.css
@@ -34,8 +34,14 @@
 
 /*-- disabled --*/
 .avonni-visual-picker-link_disabled {
-    border-color: #c9c7c5;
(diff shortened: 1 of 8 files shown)


src/modules/base/visualPickerLink/visualPickerLinkInfoOnly.css
```diff
@@ -31,3 +31,15 @@
  */
 
 @import './shared.css';
+
+/*-- disabled --*/
+.avonni-visual-picker-link_disabled {
+    border-color: #c9c7c5;
+    background-color: #ecebea;
```
nawel-aa, line 38: I think styling hooks specific to the disabled mode should be created for the background and border colours. Otherwise it is impossible to customize them.

nawel-aa (approved):
Everything seems good to go! 👍

jbverger (approved):

PR #616, opened 2023-06-27
[fix] (ABC-992): fixed compare img height

Jusolijus17:
### Fixes
**Image**
- Compare slider

### DIFF
--- a/src/modules/base/image/image.js
+++ b/src/modules/base/image/image.js
@@ -629,24 +629,38 @@ export default class Image extends LightningElement {
             styleProperties['max-height'] = null;
         }
 
-        if (this._height && !this._width) {
-            styleProperties.width = this._height
-                ? `${
-                      (parseFloat(this._height) / this._imgElementHeight) *
-                      this._imgElementWidth
-                  }px`
-                : this._imgElementWidth;
+        if (
+            this._height &&
+            !this._width &&
+            this.compareAttributes.orientation === 'horizontal'
+        ) {
+            styleProperties.width = `${
+                (parseFloat(this._height) / this._imgElementHeight) *
+                this._imgElementWidth
+            }px`;
+        } else if (
+            this._height &&
+            !this._width &&
+            this.compareAttributes.orientation === 'vertical'
+        ) {
+            styleProperties.width = 'inherit';
         } else if (this._width) {
             styleProperties.width = this._width;
         } else {
             styleProperties.width = `${this._imgElementWidth}px`;
         }
 
-        if (this._height && this._width) {
+        if (
+            (this._height && this._width) ||
+            (this.compareAttributes.orientation === 'vertical' && this._height)
+        ) {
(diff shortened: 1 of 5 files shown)


src/modules/base/image/image.js
```diff
@@ -598,6 +600,79 @@ export default class Image extends LightningElement {
...
+    get computedCompareImgContainerStyle() {
+        return this.compareAttributes.orientation === 'horizontal'
+            ? 'width: 50%; height: 100%;'
+            : 'width: 100%; height: 50%;';
+    }
+
+    get computedCompareImgStyle() {
+        const styleProperties = {};
+
+        styleProperties['object-fit'] = this.cropFit ? this.cropFit : null;
```
nawel-aa, line 612: 1. Do you _need_ to set the properties to `null` or could you juste set them to whatever falsy value the property contains (`undefined`, `null`, `''`)?
2. Your choice to decide if it is easier to read or not, but these ternary operator lines can be simplified into `styleProperties['object-fit'] = this.cropFit || null;`

src/modules/base/image/image.js
```diff
@@ -598,6 +600,79 @@ export default class Image extends LightningElement {
...
+            styleProperties['min-width'] = this._width ? this._width : null;
+            styleProperties['max-width'] = this._width ? this._width : null;
+            styleProperties['min-height'] = this._height ? this._height : null;
+            styleProperties['max-height'] = this._height ? this._height : null;
+        } else {
+            styleProperties['min-width'] = null;
+            styleProperties['max-width'] = null;
+            styleProperties['min-height'] = null;
+            styleProperties['max-height'] = null;
+        }
```
nawel-aa, line 630: 1. You can definitely refactor this big block into one line for each property. Something like:
`styleProperties['min-width'] = this.staticImages && this.width ? this.width : null;`
2. If there is a public property (`this.width`), we use it instead of the private property (`this._width`). Unless we want to set it, of course, then we set the private property. It makes it clearer that the public property exists, and it is in case there is some computation done in the `get propertyName()`.

src/modules/base/image/image.js
```diff
@@ -744,6 +819,18 @@ export default class Image extends LightningElement {
             .toString();
     }
 
+    get computedSliderHandleStyle() {
+        return this.compareAttributes.orientation === 'horizontal'
+            ? 'flex-direction: row;'
+            : 'flex-direction: column;';
```
nawel-aa, line 825: You can use an [SLDS class](https://www.lightningdesignsystem.com/utilities/grid/#Top-to-Bottom) toggle for this, instead of a custom style.

nawel-aa (changes requested):

nawel-aa (commented):
I forgot to mention it, but one test is also broken. :)

nawel-aa (approved):
🎉

jbverger (approved):

PR #621, opened 2023-07-03
[fix] (ABC-998): fixed thumbnail issue with cmp img

Jusolijus17:
### Fixes
**Image**
- Fix thumbnail not applied to compare image
- Fix border image radius not working

### DIFF
--- a/src/modules/base/image/image.css
+++ b/src/modules/base/image/image.css
@@ -44,11 +44,10 @@
     padding: 0.25rem;
     background-color: #fff;
     border: 1px solid #dee2e6;
-    border-radius: var(--avonni-image-radius-border, 0.25rem) !important;
 }
 
-.avonni-image_magnifier_thumbnail {
-    margin: 0.25rem;
+.avonni-image_container.avonni-image_thumbnail {
+    border-radius: var(--avonni-image-radius-border, 0.25rem);
 }
 
 .avonni-image_fit-content {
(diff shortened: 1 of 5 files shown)


src/modules/base/image/image.css
```diff
@@ -68,6 +68,7 @@
 .avonni-image_container {
     position: relative;
     width: fit-content;
+    overflow: auto;
```
nawel-aa, line 71: This overflow causes the magnifier to be hidden. If you really need it, maybe you should apply it to another `div`?

https://github.com/avonni/base-components/assets/71400095/ddb14088-e917-4faa-b221-e4801c57745e

nawel-aa (changes requested):

nawel-aa (changes requested):
You reference the thumbnail padding at many places in your code (`0.25rem` or `0.5rem`). It is a bit of a problem if we want to change it one day, and it makes the code more complicated for one small feature.
I think it would be simpler to add another wrapping `<div>`, inside the current container `<div>`. The only purpose of this nested `<div>` would be to have the `slds-is-relative` class, which would make sure that when the thumbnail border is visible, the images are still positioned inside the border and not over it.

src/modules/base/image/magnifier.js
nawel-aa, line 1: These three comments should be removed.

nawel-aa (changes requested):

nawel-aa (approved):

PR #625, opened 2023-07-11
[fix] (ABC-1002): removed background when completed

Jusolijus17:
### Fixes
**Visual Picker Link**
- Removed background when completed

### DIFF
--- a/src/modules/base/visualPickerLink/visualPickerLink.js
+++ b/src/modules/base/visualPickerLink/visualPickerLink.js
@@ -187,6 +187,7 @@ export default class VisualPickerLink extends LightningElement {
         return classSet('avonni-visual-picker-link__tile')
             .add({
                 'slds-welcome-mat__tile_complete': this._completed,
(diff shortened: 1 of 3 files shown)


src/modules/base/visualPickerLink/visualPickerLink.js
```diff
@@ -187,6 +187,7 @@ export default class VisualPickerLink extends LightningElement {
         return classSet('avonni-visual-picker-link__tile')
             .add({
                 'slds-welcome-mat__tile_complete': this._completed,
+                'avonni-visual-picker-link__tile_complete': this._completed,
```
nawel-aa, line 190: If both classes are applied on the same conditions (`completed` is true), you should add them together in one line:
```
'slds-welcome-mat__tile_complete avonni-visual-picker-link__tile_complete': this.completed
```

nawel-aa (changes requested):

nawel-aa (approved):
👌

PR #651, opened 2023-08-10
[feat](ABC-1030): Create Avonni Focus Trap component

Hubert06:


src/modules/base/focusTrap/__tests__/focusTrap.test.js
nawel-aa, line 1: I think it's fairly doable to add tests. :)

lwc.config.json
```diff
@@ -174,6 +174,10 @@
             "name": "c/filterMenu",
             "path": "src/modules/base/filterMenu/filterMenu.js"
         },
+        {
+            "name": "c/focusTrap",
+            "path": "src/modules/base/focusTrap/focusTrap.js"
+        },
```
nawel-aa, line 180: To make the component available publicly, you need to add the same lines with `avonni/focusTrap`, and add it to the `expose `array at the end of the file.

src/modules/base/focusTrap/__requirements__/requirements.md
nawel-aa, line 1: If we want this file to appear in [our documentation](https://www.avonnicomponents.com/), it needs to be placed in a `__docs__` folder, and renamed `focusTrap.md`.
A level 3 title should also be added on the first line (for example `### Requirements`), for more consistency.

src/modules/base/focusTrap/focusTrap.js
```diff
@@ -0,0 +1,92 @@
+import { api, LightningElement } from 'lwc';
+import { findAllTabbableElements } from './focusUtils';
+
```
nawel-aa, line 3: To make the component appear in [our documentation](https://www.avonnicomponents.com/), you need to add a [class JSDoc comment](https://avonnicreator.atlassian.net/wiki/spaces/LDG/pages/1888616449/Javascript+documentation+JSDoc#Allowed-block-tags-for-classes) with a `@public` tag. In your case, you won't have a `@storyId`, since there is no storybook example for this component.

src/modules/base/focusTrap/focusTrap.js
```diff
@@ -0,0 +1,92 @@
...
+            this._initialized = true;
+
+            this._startNode = this.template.querySelector('[data-start]');
+            this._endNode = this.template.querySelector('[data-end]');
+        }
+    }
+
+    /**
+     * Focuses the first focusable element in the focus trap.
+     */
```
nawel-aa, line 23: This JSDoc comment needs to have a `@public` tag for the method to appear in our doc.

src/modules/base/focusTrap/focusTrap.js
```diff
@@ -0,0 +1,92 @@
...
+    _handleFocusOut() {
+        // This assumes that a focusin will be dispatched after a focusout
+        this._pendingFocusOut = true;
+        // eslint-disable-next-line @lwc/lwc/no-async-operation
+        requestAnimationFrame(() => {
+            if (this._pendingFocusOut) {
+                this._focused = false;
+            }
+        });
+    }
```
nawel-aa, line 53: Our conventions are all mixed up now that we are working on several repos. 😊
But you probably should follow one of those two conventions when naming your methods and properties:
- The **base components** convention, where we never use a `_` in the name of a method, and use a `_` in the name of a property only if it is not called in the HTML.
- The **flow screen components** convention, where we use a `_` in the name of the methods and properties that are not called in the HTML.

nawel-aa (changes requested):
My comments mostly concern the fact that the component won't be publicly available in its current state. Aside from that it works well in all the base components that use it.

src/modules/base/focusTrap/__tests__/focusTrap.test.js
Hubert06: I added some tests. Could not figure out how to add elements inside the slot in the tests, so they are really basic and just for good measure.

nawel-aa (approved):
All good! 👌

jbverger (approved):

jbverger (approved):

PR #658, opened 2023-08-17
(ABC-1038): Avatar Group responsiveness

Jusolijus17:
### Features
**Avatar Group**
- Display "+..." when overflow

### Fixes
**Avatar Group**
- Responsiveness for tags not working properly when click on "Show More"

### DIFF
--- a/src/modules/base/avatarGroup/avatarGroup.js
+++ b/src/modules/base/avatarGroup/avatarGroup.js
@@ -33,6 +33,7 @@
 import { LightningElement, api } from 'lwc';
 import { classSet, generateUUID } from 'c/utils';
 import { keyCodes, normalizeString, normalizeArray } from 'c/utilsPrivate';
+import { AvonniResizeObserver } from 'c/resizeObserver';
 
 const AVATAR_GROUP_SIZES = {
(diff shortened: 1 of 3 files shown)


src/modules/base/avatarGroup/avatarGroup.js
```diff
@@ -742,6 +839,47 @@ export default class AvatarGroup extends LightningElement {
...
+
+        if (this.actionButtonElement) {
+            const actionButtonStyles = window.getComputedStyle(
+                this.actionButtonElement
+            );
+            const actionButtonWidth =
+                parseFloat(actionButtonStyles.width) +
+                parseFloat(actionButtonStyles.marginLeft) +
+                parseFloat(actionButtonStyles.marginRight);
+            availableWidth -= actionButtonWidth;
```
nawel-aa, line 864: I think you can use `this.actionButtonElement.offsetWidth` instead of using `getComputedStyle()`, can't you? 🤔

nawel-aa (commented):
It's working as is, but unless there is a specific case I didn't think of, I think the code is overly complicated for no reason. 🙂

nawel-aa (approved):

PR #1026, opened 2025-06-18
[Feat] (ABC-1423): Tree - Add attributes rootSlottableTypes, noSlots, unselectable, hiddenActions, type and slottableTypes

OliBigo:
### Features
**Tree**
- Added attribute `root-slottable-types`
- Added attributes to `items`: 
   - `hiddenActions`
   - `noSlots`
   - `slottableTypes`
   - `type`
   - `unselectable`

### DIFF
--- a/src/modules/base/primitiveTreeItem/primitiveTreeItem.js
+++ b/src/modules/base/primitiveTreeItem/primitiveTreeItem.js
@@ -646,6 +646,15 @@ export default class PrimitiveTreeItem extends LightningElement {
     }
 
     /**
+     * True if the item has actions and should not hide them.
+     *
+     * @type {boolean}
+     */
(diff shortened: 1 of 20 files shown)


src/modules/base/primitiveTreeItem/primitiveTreeItem.js
```diff
@@ -77,6 +77,7 @@ export default class PrimitiveTreeItem extends LightningElement {
     _enableInfiniteLoading = false;
     _expanded = false;
     _fields = [];
+    _hiddenActions = false;
```
Hubert06, line 80: hiddenActions should be an array of action names to hide. Otherwise it would be called hideActions

Hubert06 (changes requested):

PR #1043, opened 2025-08-07
[Feat](ABC-1445): Avatar - Add Icon Position Attribute

lyg2:
### Features
**Avatar**
- Added the `iconPosition` attribute.

### DIFF
--- a/src/modules/base/avatar/__docs__/avatar.stories.js
+++ b/src/modules/base/avatar/__docs__/avatar.stories.js
@@ -148,6 +148,18 @@ export default {
                 category: 'Details'
             }
         },
+        iconPosition: {
+            name: 'icon-position',
+            control: {
+                type: 'select'
+            },
+            options: ['start', 'center', 'end'],
+            description: 'Position of the avatar icon.',
+            table: {
+                defaultValue: { summary: 'start' },
+                type: { summary: 'string' }
+            }
+        },
         initials: {
             control: {
                 type: 'text'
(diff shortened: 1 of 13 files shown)


src/modules/base/avatar/avatar.js
```diff
@@ -284,6 +289,24 @@ export default class Avatar extends LightningElement {
         this._hideAvatarDetails = normalizeBoolean(value);
     }
 
+    /**
+     * The position of the avatar icon. Valid values are top, center, end.
```
renaudaf, line 293: The valid values are not the same as the definition in stories.js (start vs top)

src/modules/base/avatar/avatar.js
```diff
@@ -284,6 +289,24 @@ export default class Avatar extends LightningElement {
...
+     *
+     * @public
+     * @type {string}
+     * @default top
```
renaudaf, line 297: review default value

src/modules/base/avatar/__docs__/avatar.stories.js
```diff
@@ -148,6 +148,18 @@ export default {
                 category: 'Details'
             }
         },
+        iconPosition: {
```
renaudaf, line 151: the default value needs to be added to the args line 368

renaudaf: The default value is still missing from the args

renaudaf (changes requested):

renaudaf (approved):

